Issue #9, opened 2022-10-08
When SideMenu HeaderContent is empty, Menu Items can conflict with Open/Close button 


DmitryGolubenkov:
## Steps to reproduce
1. Create an Avalonia application and import the library
2. Create `DesktopPage` component in MainWindow
3. Create `SideMenu` component inside, following instructions from readme
4. Leave `HeaderContent` empty or remove it, add new `MenuItems` with icons or just with long text (not necessary)

### Result: 

The open/close sidebar button and the sidebar items get placed on the same place. Also, the sidebar button becomes unusable.
<details>
<summary>
Screenshot
</summary>

![image](https://user-images.githubusercontent.com/31821994/194710801-256abf05-b9f1-4908-a50c-e3aec2479482.png)  
![image](https://user-images.githubusercontent.com/31821994/194710816-674ae972-642e-461e-b50f-c6a70adc253d.png)  

</details>

### Workaround with current library version
Add an empty element to header. Example:
```
<suki:SideMenuModel.HeaderContent>
      <StackPanel Height="40"></StackPanel>
</suki:SideMenuModel.HeaderContent>
```
This should move the rest of elements under the button.

DmitryGolubenkov:
The issue also occurs if the header is present but the contents are too small to move menu items down
![image](https://user-images.githubusercontent.com/31821994/194711946-a4442233-2779-4c13-bb93-e89f00d565c2.png)

I'll continue testing and then provide a fix to this issue

DmitryGolubenkov:
![image](https://user-images.githubusercontent.com/31821994/194712151-9d4424c3-70b9-457f-8e2f-2419cf52218c.png)

PR #10, opened 2022-10-08
Fix header overlap

DmitryGolubenkov:
# Description
Fixes bug, described in #9. 

# Changes
- Added new property `HeaderContentOverlapsToggleSidebarButton` that controls if toggle sidebar button can overlap with header content.
- Fixed menu items and toggle button overlap by addin MinHeight property to Header content. Its value depends on `HeaderContentOverlapsToggleSidebarButton` value.
- Fixed header content overlappling with sidebar button. Button position now depends on `HeaderContentOverlapsToggleSidebarButton` property - if it is `True` - the original button outside DockPanel is shown. If `False` - a copied button that is inside DockPanel is shown.  

<details>  
<summary>Screenshots</summary>  

`HeaderContentOverlapsToggleSidebarButton=True`

image  

`HeaderContentOverlapsToggleSidebarButton=False`  

image

</details>  

# Notes
As the button code was copied, DRY principle was violated. But making the button a separate component could be a bad idea either.
The other problem is that I don't know what the next version of SukiUI based on Avalonia 0.11 mentioned in readme will contain, because there is no branch on GitHub where people can see the code. Because of that - this could contain breaking changes for the future release, if `Sidebar` component was modified.
And maybe the issue was already fixed in new version.

Also - maybe consider moving the documentation to Github Wiki that can be enabled in repository settings? `Readme.md` is hard to read and modify. It should show off the possibilities and list features, tell about contributing and license, but all the details, examples of implementation etc should be on their own pages on something like wiki.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,6 +121,8 @@ https://user-images.githubusercontent.com/19242427/187209227-03598bf8-c958-4577-
 
 <img src="https://raw.githubusercontent.com/kikipoulet/SukiUI/main/Images/SideMenu3.gif"></img>
 
+- Header content can overlap sidebar toggle button if its content is too wide. To prevent that, either design header content layout to be narrower, or set `SideMenuModel.HeaderContentOverlapsToggleSidebarButton` property to `False`. Default is `True`.  `HeaderContentOverlapsToggleSidebarButton` moves header content under the button.  
+
 Xaml Code Method
 </br>
 <details>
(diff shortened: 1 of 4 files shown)


DmitryGolubenkov:
@kikipoulet can you please add `hacktoberfest-accepted' label to one of the PRs that I have made? I'm one PR away from event target

PR #14, opened 2023-01-05
Binding compatibility

nixtar:
This PR will track the work/discussion on adding binding compatibility to all controls where appropriate.

Controls:
- [x] MobileNumberPicker
- [x] PercentProgressBar

Already in main:
- [x] MobileNumericUpDown
- [x] Stepper
- [x] TouchKeyboard
- [x] TouchNumericPad

### DIFF
--- a/SukiUI/Controls/MobileNumberPicker/MobileNumberPicker.axaml.cs
+++ b/SukiUI/Controls/MobileNumberPicker/MobileNumberPicker.axaml.cs
@@ -19,8 +19,6 @@ public partial class MobileNumberPicker : UserControl
     {
         AvaloniaXamlLoader.Load(this);
     }
-
-    
     private int _value;
     
     /// <summary>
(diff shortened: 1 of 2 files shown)


nixtar:
If you are happy with the style of my MobileNumberPicker changes then I'll go ahead and update the PR with PercentProgressBar also.

kikipoulet:
It is absolutely perfect !

nixtar:
Updated PercentProgressBar.

Also added some caching for the controls so FindControl isn't being used every time.
FindControl is a bit of a hot spot especially if you are updating the progress bar a lot. (if you are not buffering the progress bar updates on a file download for eg)

Experimented using ChatGPT to add comments to the code also 😄.

I was unable to test the changes to the control as I can't get MobileDemo or SukiTest projects to build due to some weird dependency issues. It might be because I'm using Rider. If I get time, I'd love to write a proper Intergration/Unit test for this library (never done it in Avalonia before, and never done it "properly" for other UI frameworks so it would be a sweet learning experience).

kikipoulet:
Thank you very much ! I really appreciate that clean code !

That's weird that you can't run at least SukiTest because I run it on Windows and Linux using Rider too. 
And for sure a proper Integration/Unit test would be great and appreciated if you get time :+1:

Issue #55, opened 2023-12-23
Images folder contains multiple files with the same filename.


sirdoombox:
It's a minor issue, but something I can't reasonably fix pulling the repo down in Win10 because the case sensitivity of the files is broken on contact with the FileSystem, so it'd taint any PR I'd try to make.

These are the files causing issues:
- DataGrid.gif
- ComboBox.gif
- TextBox.gif

As far as I can see some of these duplicates were meant as actual replacements, so I guess they could just be deleted, but just renaming them would be fine too.

PR #56, opened 2023-12-24
Improve title bar and rewrote WaveProgress converters.

sirdoombox:
I implemented a very simple title bar "replacement" that allows the user to interact with the title bar as expected across it's entire height (including double-click to maximise)

I also rewrote the WaveProgress converters to avoid try/catch abuse which can be a pointless performance drain if "invalid" objects are bound.

I also fixed an issue revolving around duplicate filenames mentioned in #55


sirdoombox:
I've got a few more changes cooking so I'll close this for now until I've finished them all.

kikipoulet:
Well it looks really nice thank you !

PR #57, opened 2023-12-25
Large suite of changes - Including removing dependency on Material.Icons

### LINKED ISSUE #55: Images folder contains multiple files with the same filename.

sirdoombox:
**Major Changes:**
### Completely removed dependency on `Material.Icons`/`Material.Icons.Avalonia`.
- This prevents issues where installing a different version of the package into a project would cause icons in SukiUI to be wrong.
- Completely unnecessary dependency anyway, a few mb for what turned out to be 30 icons.
- New solution involves a simple C# file with icon paths stored in StreamGeometry instances, displayed with `PathIcon`
- Now allows custom icons to be included or icons from other sources to be used without requiring another dependency.
- Provides compile time safety, and allows icons to be changed in one place and the change being consistent across the whole UI.
- Allows devs to reference these named icons themselves to improve consistency in their apps, which will also be immune to updates/changes to SukiUI.
- Unfortunately sizing these `PathIcons` is different so I simply stripped out all sizes so they can be adjusted in the future.

### Enabled nullables for the project
- This is the standard now generally and can prevent significant issues going forward, this can be unwound if it's seen to be an issue.
---
**Minor Changes:**
- WaveProgress converters rewritten to prevent try/catch abuse that can drain performance.
- Simplified Maximise/Restore window button logic, icon is now simply set by a style.
- Added optional `BusyText` property to the `BusyArea` which indicates to the user what the app is doing.
- Enabled the sort icon on `DataGrid` by giviing it a color and changed enable/disable behaviour to use `Opacity` so that space is correctly reserved for it, preventing jerky resizing behaviour.
---
**Fixes:**
- Include `.idea/` in the `.gitignore` for Rider devs.
- Stripped out all the margins/padding present in the `Button` styles as these seemed to do nothing but cause issues.
- Title bar interaction now spans it's entire height.
- Implemented a style applied on Windows only that applies an 8px padding to the window to offset an issue in Windows where windows that are maximised are over-sized to hide the system borders, but SukiUI's `NakedWindow` removes those system borders.
- Removed duplicate image files causing issues developing on Windows (Fixes #55)
---
**Outstanding Issues:**
- As stated before, when replacing all the `MaterialIcon` controls, the `Height` and `Width` properties simply had to be removed because the sizing wasn't consistent, going forward they may have to be adjusted.
- Buttons haven't been fully tested after the major changes to the padding/margins, however they _**seem**_ okay.
- If users are using Material.Icons themselves, these icons won't have their foreground set correctly by default as before, however devs can do this with a single simple style anyway.


kikipoulet:
Okay, this is really cool ! 🚀

Thank you for your contribution, these are exactly the kind of improvements the library needs.

PR #59, opened 2023-12-27
Another chunk of changes.

sirdoombox:
I've made some more big breaking changes, I see that the plan is to make the old "Glass" style the standard so I've made some changes in that direction.

***More movement towards making Glass the default***
- Removed all references to `GlassEnabled` and `GlassDisabled` as Glass is now the default.
- Moved `GlassCard` into the `SukiUI.Controls` namespace because it's now a default control.

***Dialog improvements***
- Rewrote `InteractiveContainer.ShowDialog` so now you can send it any arbitrary object as well as a `Control`, and it will attempt to locate a suitable `View` using a suitable locator if you've assigned one to the App's `DataTemplates`, a small change but necessary for VM-First development.
- Added a control that can handle clicks outside of the dialog, and request a close of the currently active dialog.
- Added a property/arg to the `ShowDialog` method enabling/disabling the ability for the dialog to be closed by simply clicking outside of it.

***Minor changes***
- Added `Foreground` property value to the `ContentPresenter` of `Button` so that `<Button Content="Hello, World!"/>` is correctly styled. (This could be made more specific if necessary)
- Collapsed the styles in `InteractiveWindow` at the bottom to improve readability.

***Outstanding issues***
- The dialog system is a little clunky and might need a rethink.
- Making the window `NakedWindow` by default is probably a smart move given the design is heavily focused around that.
- Some controls completely disappear when the tab is changed in the SukiTest app.
- Opening a dialog with the new blur animation seems to be a little sluggish (on my machine anyway).


kikipoulet:
That's so great thank you very much ! :+1: 

- I think making NakedWindow as default would be problematic for users that don't use DesktopPage, and I think there are unfortunately a lot

- The controls disappearing really annoys me since too much time, because I don't understand why it does that. Well, I guess it's because of the SideMenu page declarations that stay the same View instance between page changes instead of creating a new view everytime, but I like that easy xaml syntax and even if some MVVM guru would think it's dirty it's not supposed to do that. 

At first I had the problem with transitioningcontentcontrol (https://github.com/AvaloniaUI/Avalonia/issues/12158) but using contentcontrol solved it. Now it seems that some arbitrary controls are affected again. Annoying. 

- Maybe reducing the blur(18) of the animation to less maximum blur would make it less sluggish for different platforms, on linux it was fine. blur(18) was generous anyway

kikipoulet:
Controls disappearing definitely fixed in d3bd795 by changing the sidemenu paradigm

There is no "Content" field anymore but a "ContentType" that will instanciate a new View of the type given everytime the page is switched to

PR #77, opened 2024-01-02
Minor fixes.

sirdoombox:
***Changes*** 
- Introduced `PathIcon` styles.
- Rewrote the `ThemingViewModel` to be a little more sane.
- Very basic navigation service for the demo app so that pages can request other pages be opened.
- Rolled the background controls into `SukiTheme` so now everything visual is in one place, stops the horrendous mess of trying to get the root `SukiWindow` just to manually modify a property on that.

***Outstanding Issues***
- The current way of displaying and toggling between themes in the theming view doesn't allow for easy updates based on changes, there are `TODO` in the file marking where the issue is.
- `ShowMeTheXaml` doesn't seem to like displaying multi-line text at all, must be an issue with `TextBox`, a bit annoying.
- `TextBlock` which have a `<LineBreak/>` in them don't correctly update when the theme changes, this is however seems to be an Avalonia bug and I don't know that there's much we can do about it from our end.
  - https://github.com/AvaloniaUI/Avalonia/issues/12404

### DIFF
--- a/SukiUI.Demo/App.axaml.cs
+++ b/SukiUI.Demo/App.axaml.cs
@@ -8,6 +8,7 @@ using Avalonia.Markup.Xaml;
 using Microsoft.Extensions.DependencyInjection;
 using SukiUI.Demo.Common;
 using SukiUI.Demo.Features;
+using SukiUI.Demo.Services;
 
 namespace SukiUI.Demo;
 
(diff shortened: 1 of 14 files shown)


sirdoombox:
My username is `doombox.` all lowercase with the `.` full stop at the end. I'm not sure who that is.

I'm also available in the C# discord https://discord.gg/invite/csharp

PR #95, opened 2024-01-13
Refactor PropertyGrid

Insire:
**_Changes_**

- PropertyGrid was overhauled and now uses a MVVM and DataTemplates
- PropertyGrid fields can now be customized, by replacing the ``PropertyGridTemplateSelector`` with a custom one
- PropertyGrid ViewModels were added (one for each BCL type)
- fixed a bunch of warnings
- added an extension for searching controls that will throw if the required control wasnt found, fixing various nullable warnings
- replaced Thread.Sleep with Task.Delay
- applied best practices for async usage
- added ``IsExternalInit `` class,  which allows us to use C# 9's init and record features in older target frameworks like .NET Standard 2.0

_**Outstanding Issues**_

- PropertyGrid could support more BCL types like colors with a colorpicker for example
- not all warnings were fixed
- PropertyGridWindow has a transparent backround for some reason, didnt figure out why


kikipoulet:
Had the idea to style the propertygrid with the settingslayout control, to be able to do a complete responsive settings page with just a viewmodel 🤔

Insire:
> Had the idea to style the propertygrid with the settingslayout control, to be able to do a complete responsive settings page with just a viewmodel 🤔

Thats still possible, only have to replace the ItemsControl in the PropertyGrid with the SettingsLayout. All the other parts are public and can be used for that aswell.

PR #118, opened 2024-02-01
Add async SukiHost.ShowDialog.

yuzh0816:


sirdoombox:
I do have a proper API for this on my list of things to do for a major future version, or fork, depending on the way the wind blows. The rewrite will include a proper dialog container and likely some default templates for simple confirmation dialogs and such and a clean way to return results.

In the short term though, the use of a 1ms busy wait here is quite wasteful and it'd be better to have the dialog view/viewmodel you're using implement the callback/events in your own project. If this API specifically is really necessary now, it should use a `TaskCompletionSource` wired it up the dialog visible property changing to minimise waste.

Insire:
I dont see the need for including this in this repo, since if you really want this, it can easily be a regular extension method in your own project.

PR #130, opened 2024-02-09
Bump Avalonia to 11.0.9

sirdoombox:
This is a PR with the change and I think should serve as a discussion point about what is best for SukiUI going forward, should we continue to keep pace with Avalonia's minor versions/revisions and follow major versions with our own major version changes?  (I.E Avalonia 12 would be SukiUI 7.0). Will failing to keep up with Avalonia's minor versions/revisions cause significant issues for users? I think it's worth discussing before making this change, because I'm not sure what to do for the best long term.

As a note, somewhere along the line Avalonia introduced a regression which causes `TextBlock` runs to have a brand new bug. This can be seen in `TextView.axaml` the offending code has been commented out, nothing I have tried seems to work to fix it even though that exact same button setup works fine in other pages.

This PR is up so it can be checked out locally and probably shouldn't be merged until a decision is made.

### DIFF
--- a/SukiUI.Demo/Features/ControlsLibrary/ProgressView.axaml
+++ b/SukiUI.Demo/Features/ControlsLibrary/ProgressView.axaml
@@ -148,7 +148,7 @@
 
                 <controls:GlassCard Width="500">
                     <controls:GroupBox Header="Stepper">
-                        <StackPanel  Spacing="15">
+                        <StackPanel Spacing="15">
                             <showMeTheXaml:XamlDisplay Margin="0,20,0,0" UniqueId="Stepper">
                                 <controls:Stepper Index="{Binding StepIndex}" Steps="{Binding Steps}" />
                             </showMeTheXaml:XamlDisplay>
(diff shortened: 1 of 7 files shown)


kikipoulet:
That's a good question, the trap is that changes between minor versions often break things, but we can try to follow the minor Avalonia releases with ours, or simply do it everytime SukiUI works perfectly with it and ignoring avalonia minor releases that could cause problem like the text bug.

I think following Avalonia major version for our major version is a good idea, 6.0 is for Avalonia 11, and 7.0 for Avalonia 12.
Or jumping directly to 11 and 12 for more clarity ?

sirdoombox:
Might be necessary to evaluate every Avalonia release in some way if there are some important fixes we really need, although I think what makes the most sense is supporting whatever the default Avalonia version is for the most up-to-date templates are. This way we provide the best out of the box experience because that's what most devs will likely end up using.

<https://github.com/AvaloniaUI/avalonia-dotnet-templates/blob/e4a489ae828f005f625145c563785174403e267c/templates/csharp/app/.template.config/template.json#L59>

The templates actually define a constant we could just keep an eye on over time.

The only thing that would be nice is some obvious way for devs to know which version of Avalonia goes with which version of SukiUI.

kikipoulet:
I think it is indeed the good balance. Keeping up with the templates to not necesseraly jump into every minor release, but keeping an eyes on it for breaking changes and interesting fixes.

+1 On keeping the Demo on the most recent version, so we can quickly see if an update for SukiUI is needed

sirdoombox:
It seems Avalonia have fixed some of the bugs introduced in `11.0.9` but those fixes won't be coming for a while.

I'll leave the project versions as-is for now until Avalonia release something that's actually stable enough for us to even bump the Demo app.

PR #143, opened 2024-02-25
Fix bug of window titlebar.

yuzh0816:


sirdoombox:
Uh, I'm not quite sure how the associated PR fixes this bug, or even that there is a bug at all. Setting `CanResize` to false correctly prevents the window from being resized in the demo application.

All this PR does is creates a reversion in behaviour for the title bar when the window is fullscreen and the titlebar is still visible. The behaviour is very wonky and the window doesn't behave properly if the window is allowed to be in the "unlocked" state when it is in fullscreen mode.

PR #173, opened 2024-04-28
PropertyGrid supports sorting

YJammak:
Add PropertyIndexAttribute to enable PropertyGrid to support sorting features

### DIFF
--- a/SukiUI/Controls/PropertyGrid/InstanceViewModel.cs
+++ b/SukiUI/Controls/PropertyGrid/InstanceViewModel.cs
@@ -42,6 +42,17 @@ namespace SukiUI.Controls
             return null;
         }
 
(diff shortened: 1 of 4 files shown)


Insire:
I'm not sure this is a great addition to SukiUI. Up until now, the PropertyGrid uses the existing types of .NET and mimicks the behavior of the original implementation of the winforms property grid. Also, anyone who needs  a custom sorting solution, can just subclass the ``InstanceViewModel`` and override the result provided by ``InstanceViewModel.GenerateCategories``.

Finally, introducing this property will lead ppl to breaking MVVM, by taking a dependency in their ViewModels or even Models on SukiUI.

YJammak:
Thanks, I didn't realize that sorting could be achieved using a subclass of InstanceViewModel.

PR #175, opened 2024-05-02
UI: Better looking ToggleButton rotation in TreeViewItem

AuroraZiling:
### Behavior
- Before
before

- After
after

### Cause

The center point of the original SVG image is to the upper left, but in order not to modify the original design of Material Design, only the RenderTransformOrigin is modified.

image


kikipoulet:
Great PR thanks ☺️

PR #183, opened 2024-05-12
Visual Optimization of `SukiSideMenu`

AuroraZiling:
## Description

This PR will result in two changes:
  - Able to set whether the movement animation of `SukiSideMenuItem` content is enabled or not
  - `SukiSideMenuItem` content movement using non-linear animation

Current Animation:
before

### First Change

I added a new property to `SukiSideMenu`: `IsSelectedItemContentMovable` (Default value = true)

This property implies whether or not the content of the Item triggers a movement animation when it is selected (Padding)

For each `SukiSideMenuItem`, the property is `IsContentMovable`

When `IsSelectedItemContentMovable` is false, the behavior is as follows:

after

### Second Change

When `IsSelectedItemContentMovable` of `SukiSideMenu` is true, movement animation of `Item` is nonlinear (CubicEaseInOut)

after2

### DIFF
--- a/SukiUI/Controls/SukiSideMenu.axaml.cs
+++ b/SukiUI/Controls/SukiSideMenu.axaml.cs
@@ -24,7 +24,14 @@ public class SukiSideMenu : SelectingItemsControl
         set => SetValue(IsMenuExpandedProperty, value);
     }
 
- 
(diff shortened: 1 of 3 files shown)


kikipoulet:
Super nice 👍

Issue #184, opened 2024-05-13
`LICENSE` and cute `suki_photo.ico` are included in the project by default?


AuroraZiling:
After creating a new project with `SukiUI`:
![image](https://github.com/kikipoulet/SukiUI/assets/69903523/62eacd7f-b690-46f5-b953-f232025cbf57)

Related code:

[SukiUI.csproj#L53](https://github.com/kikipoulet/SukiUI/blob/cfb3db2c6f9c0d69799a0f6b526ee0b80ea5fcd2/SukiUI/SukiUI.csproj#L53)

[SukiUI.csproj#L57](https://github.com/kikipoulet/SukiUI/blob/cfb3db2c6f9c0d69799a0f6b526ee0b80ea5fcd2/SukiUI/SukiUI.csproj#L57)

PR #185, opened 2024-05-13
Fix and more kinds of toasts

AuroraZiling:
## First Change

Issue #184 

I don't know if the attributes of these two files are set this way on purpose or not.

If it was on purpose, I would have withdrawn the change.

## Second Change

Support 4 kinds of toast with different icon and color *Info, Success, Warning and Error*

Screenshot:

effect

Concern:

I'm still learning Avalonia, so I'm not sure if there's a better way to implement this code.

image


kikipoulet:
This is really cool don't worry !

PR #205, opened 2024-06-01
[Feat] Badge

AuroraZiling:
custom properties:
- Header *string*
- Appearance *NotificationType*
- IsOpaque

effect:

badge


kikipoulet:
A badge control is not a bad idea but to be more useful, I think it should be a control that could be shown on another like this :



Otherwise it multiply controls for something people's can do themselves too easily 🤔

AuroraZiling:
i think that could be InfoBadge

these changes are indeed not so useful :(

PR #240, opened 2024-07-02
Introduce new colours and fix up buttons.

sirdoombox:
### Minor Changes
- Introduced a `SukiPrimaryDarkColor` and `SukiAccentDarkColor` which is currently pre-calculated in `SukiColorTheme` using a scalar to be 50% "darker".
- These colours have been used to update the disabled styling on buttons.
  - Using `Opacity` was undesirable as making buttons partially transparent looked odd in many cases.
- Tweaked the app icon in the `SukiSideMenu` in the demo app to have visible pointerover interaction.
  - I also set the `ToolTip.ShowDelay` to `0` on this temporarily as a means of keeping an eye on a bug in Avalonia which causes tooltips to flicker in some cases. 

### Outstanding Issues
- `CanResize` bug still plagues us but seems to be an Avalonia issue.
- `ToolTip.Tip` bug is in a similar spot.
- Button styling isn't _perfect_ and bits and pieces might need tweaking.
- The exact value of the dark colour isn't final and can be tweaked.
- It might be an idea to run through and see where we can use the `Dark` variants of colours instead of numbered `Opacity` variants to prevent transparency being introduced where it shouldn't.

### DIFF
--- a/SukiUI.Demo/SukiUIDemoView.axaml
+++ b/SukiUI.Demo/SukiUIDemoView.axaml
@@ -93,6 +93,18 @@
         </MenuItem>
     </suki:SukiWindow.MenuItems>
     <suki:SukiSideMenu ItemsSource="{Binding DemoPages}" SelectedItem="{Binding ActivePage}">
+        <suki:SukiSideMenu.Styles>
+            <Style Selector="Image.AppIcon">
+                <Setter Property="Transitions">
+                    <Transitions>
+                        <DoubleTransition Property="Opacity" Duration="0.1" />
+                    </Transitions>
+                </Setter>
+                <Style Selector="^:pointerover">
+                    <Setter Property="Opacity" Value="0.5" />
+                </Style>
+            </Style>
+        </suki:SukiSideMenu.Styles>
         <suki:SukiSideMenu.ItemTemplate>
             <DataTemplate>
                 <suki:SukiSideMenuItem Header="{Binding DisplayName}">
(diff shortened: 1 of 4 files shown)


kikipoulet:
About the button styling not perfect, what are your remarks ? 🤔

I had the thought recently that cornerradius could be better, even thinking that it would be better to have a custom implementation based on a quadratic Bezier curve instead of a "dumb" circle.



A little off-topic, just a thought that I had so I put it here 👀

sirdoombox:
My main remarks on the button styling as it is is animations being a little inconsistent and flashy in places and the styling classes being quite complex and quite tough to maintain. Also the loading icon doesn't have it's colour set properly which I forgot to do. I think we're offering a good range of options though so I don't think it is worth trying to reduce complexity by reducing options. Maybe there is a better way to organise the classes to help with bloat.

As for changing rounded corner rendering, I don't know enough about the API available to do this with simple styles and how expensive it might be. I think we'd have to look into how Avalonia implements the rounded corners internally to see if it's easy to "override" in some way.

sirdoombox:
To add to this as I just thought of it, we might need `SukiPrimaryLightColor` which is 50% _lighter_ for use in light mode? I think if I'm not mistaken in light mode disabled controls usually tend to get lighter, not darker?

kikipoulet:
> To add to this as I just thought of it, we might need `SukiPrimaryLightColor` which is 50% _lighter_ for use in light mode? I think if I'm not mistaken in light mode disabled controls usually tend to get lighter, not darker?

Yes I think so, lighter to be more discrete

sirdoombox:
I did try to introduce a lighter colour for disabled buttons but it doesn't look right at all. I really do hate the light theme...

PR #255, opened 2024-07-22
Migrate to Avalonia 11.1

sirdoombox:
### Changes
- Migrate to Avalonia 11.1.0 (Release)
- Fix a binding error in TextBox styles (I only managed to catch this when I was implementing logging in my own project, I might add some Avalonia logging to the demo app so we can see what's going on internally a bit more easily)
- Switch every inline button in the demo app to `HyperlinkButton`
  - Also removed the `Hyperlink` button style as it's redundant now.
- Create a basic style for `HyperlinkButton`

### Outstanding issues
- ~~HyperlinkButton brush transition not really working, though that might be preferable.~~
- The annoying text run not updating the text color bug is still present in Avalonia requiring the annoying workaround still...
- The tooltip bug still remains, I have to assume it isn't a bug in Avalonia but a feature and we have to adjust our implementation somehow?
- The `ToggleSwitch` style is still broken in the settings flyout, weird state bug that causes it to flicker whenever the flyout is opened

### DIFF
--- a/SukiUI.Demo/Features/ControlsLibrary/Dialogs/DialogsView.axaml
+++ b/SukiUI.Demo/Features/ControlsLibrary/Dialogs/DialogsView.axaml
@@ -19,9 +19,8 @@
                     <TextBlock>
                         As the API for this resides entirely in C#, the source for the examples can be viewed on GitHub at the URL below.
                     </TextBlock>
-                    <Button Classes="Hyperlink"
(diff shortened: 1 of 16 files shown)


AuroraZiling:
For issue 1, adding these may work:
image

sirdoombox:
Ah yes, silly mistake. Thanks!

PR #261, opened 2024-08-06
Use DefaultDemiBold to set the DemiBold in the style to avoid blurring caused by being too thick in some fonts

YJammak:
Use `DefaultDemiBold` to set the `DemiBold` in the style to avoid blurring caused by being too thick in some fonts.


kikipoulet:
This is an unexpectedly good idea

Issue #266, opened 2024-08-10
[Feat] Implement `SukiToast` based on `Avalonia.Controls.Notifications`


AuroraZiling:
### Use the search function to see if this feature has already been suggested.

- [X] I have searched for duplicate issues.

### Description of new feature

Since Avalonia provides Notification, I thought it would be better to implement a custom Toast based on that.

One more thing is that doing this would solve the weird issues of Dispatcher and Task in the toasts that occur with probability.

### Package Version

6.0.0-beta7

sirdoombox:
I'll look into this. I think maintaining an entirely custom implementation of toasts is going to get to be a headache long-term and relying on first-party tooling would be ideal.

The only downside I see at the moment is that it's not that actively maintained, with the last update being 6 months ago so there might be some risk in doing a bunch of work on something that might well be abandoned - similar issue with TreeDataGrid which is not maintained at all and is implemented quite poorly.

kikipoulet:
> I think that this is worth looking into, I'll take a look if I get the chance in the next couple days. I think maintaining our own  entirely custom implementation of toasts is going to get to be a headache long-term and relying on first-party tooling would be ideal.
> 
> The only downside I see at the moment is that it's not that actively maintained, with the last update being 6 months ago so there might be some risk in doing a bunch of work on something that might well be abandoned - similar issue with TreeDataGrid which is not maintained at all and is implemented quite poorly.

If I remember well, we actually made the toasts because of it. 

We have the file NotificationStyle and NotificationCardStyle in the Theme folder that styled it, but the Avalonia notification system was a headache to use and not really consistent compared to a simple called to SukiHost. 

Moreover it let us control the API and do more complex things like the "Update" notification with a progressbar in SukiDemo. If I remember well, such thing is impossible with Avalonia system.

sirdoombox:
I never really investigated `Avalonia.Notifications` that deeply. I really like the MVVM architecture they have from my first look at it which makes localising a notification to a specific window, or area or even inside a control a _lot_ cleaner. It also makes the entire thing optional and gives the end-user a lot more control. It is over-complicated internally for what we need and it lacks other things that we require which would mean extending it quite significantly I think.

[...]

PR #270, opened 2024-08-12
Hosts rewrite.

sirdoombox:
### Major Changes
- Complete rewrite of the SukiHost API - Styling remains largely unchanged.
- `SukiHost` is no more. That single object was quickly turning into a nightmare.
- Add a `Hosts` property to `SukiWindow` so people can choose which hosts they do or don't want and implement their own window-level host style controls as needed.
- Tore out and rewrote the toast API as it's own separate entity.
  - Now loosely mirrors the `Avalonia.Notification` API, for people familiar with that this will be far easier to pick up.
  - Much more MVVM friendly, no more horrible static state and needing to find specific `Window` instances.
  - Allows a `SukiToastHost` to be used anywhere or nowhere, you can now have toasts inside a button if you want.
  - Uses a builder pattern for constructing toasts.
    - This is far easier to expand and modify the API without causing massive headaches, no more adding extra arguments and overloads to `SukiHost.ShowToast`.
  - You can now have any number of `ActionButtons` within a toast.
  - Much more fine grained control over the behaviour of toasts - Try out the `Action Toasts` in the demo of this PR.
  - No more async/tasks, not seeing any blocking going on so this is unlikely to need to change.
  - Toast location now works correctly and is set by the host itself. You can change this at runtime, and because you can have multiple toast hosts at once in the window you can have one in each corner and show toasts in all 4 locations at once if you wanted to.
- Tore out and rewrote the dialog API as it's own separate entity 
  - API is very similar to the toasts, so it should be fairly self explanatory how to use it.
  - Creating simple confirmation/multi option dialogs is now super easy, it shares the ActionButtons API with toasts.
  - Should be very simple to expand with any number of additional features, things like input fields.
  - Supports passing in a `ViewModel` and then it just renders that only, so any custom dialog is possible without having to jump through hoops.
  - No more async/tasks in the API, no blocking here either.
  - MessageBox style element has been included as part of this API.
- Fixed the background transparency animation for dialogs.

### Outstanding Issues
- Documentation is fairly basic at the moment, I've written the entirety of the page for toasts but wanted to check it was formatted correctly and provided enough detail before I work on the dialogs page.
- I can't seem to figure out how to limit the number of toasts properly. The limit sort of works, but if you spam them quick enough it's possible to exceed the limit.
- Icon appearance isn't great for dialogs, I just slapped it together so it's working, it could probably be positioned a bit more sensibly.
- The PropertyGrid stuff which used SukiHost for displaying it in a dialog is currently not working because there is no longer a static instance, this might not even need to be fixed, you can probably just slap a PropertyGrid inside a dialog yourself.
- This is obviously going to need a fresh pair of eyes and some additional testing to make sure it's solid.

PR Addresses: #266


kikipoulet:
The action buttons seems to mix between toasts sometimes :

{D1FB9B73-B0FF-4F99-95E5-0D290FDC79F6}

Also, we need a way, a "shortcut", to quickly use toasts and dialogs in the current window without MVVM for guys who just want to quickly display a message in a simple app. Maybe just give this kind of dirty snippet in a method

```   
 var sukiWindow = ((SukiWindow)TopLevel.GetTopLevel(((IClassicDesktopStyleApplicationLifetime)Application.Current.ApplicationLifetime).MainWindow));
        
var currentmanager = ((SukiToastHost) sukiWindow.Hosts.First(h => h is SukiToastHost)).Manager;
```

I know you won't like it ☠️


Otherwise this is great, especially calling toast like this 

{8403B4B1-683A-4870-986F-39F09F5AA847}

{CF4383B3-E080-4E52-A8FC-907F8F732D53}


This is perfection.

Don't worry about the syling I will correct it at the end.

sirdoombox:
I don't think there's a huge need to add a hacky workaround to the library, it's very easy to just implement it yourself if you don't want to bother with MVVM. I can add this workaround to the docs for people who want to have a lazy way to do it. Also this is basically how you should do it if you're not doing MVVM anyway.
```xml
<SukiWindow.Hosts>
	<SukiToastHost Name="ToastHost"/>
</SukiWindow.Hosts>
```
```cs
public class MainWindow : SukiWindow
{
	public static ISukiToastManager ToastManager = new SukiToastManager();

	public MainWindow()
	{
		ToastHost.Manager = ToastManager;
	}
}

// and then...

MainWindow.ToastManager.CreateToast()
	.Queue();
```

Button bleed is easily fixed, just an oversight with the pooling and I think I have an idea for how to fix the toast limits too so I'll get that done as well later on today when I get back.

kikipoulet:
> I don't think there's a huge need to add a hacky workaround to the library, it's very easy to just implement it yourself if you don't want to bother with MVVM. I can add this workaround to the docs for people who want to have a lazy way to do it. Also this is basically how you should do it if you're not doing MVVM anyway.
> 
> ```
> <SukiWindow.Hosts>
> 	<SukiToastHost Name="ToastHost"/>
> </SukiWindow.Hosts>
> ```
> 
> ```cs
> public class MainWindow : SukiWindow
> {
> 	public static ISukiToastManager ToastManager = new SukiToastManager();
> 
> 	public MainWindow()
> 	{
> 		ToastHost.Manager = ToastManager;
> 	}
> }
> 
> // and then...
> 
> MainWindow.ToastManager.CreateToast()
> 	.Queue();
> ```
> 
> Button bleed is easily fixed, just an oversight with the pooling and I think I have an idea for how to fix the toast limits too so I'll get that done as well later on today when I get back.

I agree, adding that to the doc will be enough 👍

sirdoombox:
This is mostly complete now I think. I believe that adding new features/styling elements to dialogs and toasts should be _much_ easier now and limit the breaknig changes introduced going forward.

You can implement whatever visual changes you need to the `SukiToast`/`SukiDialog` and add some extra methods so `SukiDialogBuilder`/`SukiToastBuilder` as well as the `FluentBuilder` for each without having to modify the signature of existing methods forcing devs to make any changes. It also means we won't end up with 300 different overloads in `SukiHost` which is where I feel that type was heading.

kikipoulet:
Had to do resolve a conflict manually but it seems to be okay

sirdoombox:
I'm not sure how you merged the PR without actually merging it, should it be closed?

PR #272, opened 2024-08-15
Fixed repeated animation when dismissing Toast.

YJammak:
Fixed repeated animation when dismissing Toast in ActionButton callback function.

### DIFF
--- a/SukiUI/Controls/SukiToast.axaml.cs
+++ b/SukiUI/Controls/SukiToast.axaml.cs
@@ -60,6 +60,8 @@ public class SukiToast : ContentControl, ISukiToast
         set => SetValue(ActionButtonsProperty, value);
     }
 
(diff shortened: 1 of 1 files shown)


kikipoulet:
@sirdoombox

YJammak:
When a delay is set to dismiss Toast, but Toast is manually closed in the ActionButton callback function before it is automatically dismissed, Toast will run the dismissing animation again.

sirdoombox:
I can't seem to reproduce this one locally in any of the demo situations but it is entirely possible that two OnDismissed events being raised could cause the dismiss animation to be called twice, it'd have to be well timed though. I think this fix is a little rough I think it'd be better to actually check the manager to see if it's been dismissed instead of duplicating state. I think that API should be expanded to handle that case anyway.

PR #286, opened 2024-08-30
SukiWindow adds TitleBarAnimationEnabled

YJammak:
When SukiWindow is started, if IsTitleBarVisible is False, the animation of closing TitleBar will be played. Add TitleBarAnimationEnabled to manually control whether to play the animation of switching TitleBar visible.


sirdoombox:
LGTM.

Issue #288, opened 2024-09-05
[Bug] JetBrains Rider cannot preview SukiWindow


imJack6:
### Check the following items

- [X] I have looked up relevant Issue

### Description of the issue

rt, when I look at the preview window after building, a white window is displayed first, and then the following error appears:
```
"C:\Program Files\dotnet\dotnet.exe" exec --runtimeconfig D:\Code\CSharp\Test\Test\bin\Debug\net8.0\Test.runtimeconfig.json --depsfile D:\Code\CSharp\Test\Test\bin\Debug\net8.0\Test.deps.json D:\NuGet\packages\avalonia\11.2.0-beta1\buildTransitive\..\tools\netstandard2.0\designer\Avalonia.Designer.HostApp.dll --transport tcp-bson://127.0.0.1:3972/ --method avalonia-remote D:\Code\CSharp\Test\Test\bin\Debug\net8.0\Test.dll
Initializing application in design mode
Obtaining AppBuilder instance from Test.Program
Sending StartDesignerSessionMessage
Fatal error. System.Runtime.InteropServices.SEHException (0x80004005): External component has thrown an exception.
Repeat 2 times:
--------------------------------
   at SkiaSharp.SkiaApi.sk_canvas_draw_rect(IntPtr, SkiaSharp.SKRect*, IntPtr)
--------------------------------
   at SkiaSharp.SKCanvas.DrawRect(SkiaSharp.SKRect, SkiaSharp.SKPaint)
   at SukiUI.Utilities.Effects.EffectBackgroundDraw.Render(SkiaSharp.SKCanvas, SkiaSharp.SKRect)
   at SukiUI.Utilities.Effects.EffectDrawBase.Render(Avalonia.Media.ImmediateDrawingContext)
   at Avalonia.Rendering.Composition.Drawing.Nodes.RenderDataCustomNode.Invoke(Avalonia.Rendering.Composition.Drawing.Nodes.RenderDataNodeRenderContext ByRef)
   at Avalonia.Rendering.Composition.Drawing.ServerCompositionRenderData.Render(Avalonia.Platform.IDrawingContextImpl)
   at Avalonia.Rendering.Composition.Server.ServerCompositionDrawListVisual.RenderCore(Avalonia.Rendering.Composition.Server.ServerVisualRenderContext, Avalonia.Platform.LtrbRect)
[...]

PR #291, opened 2024-09-12
Implement software rendering fallback.

sirdoombox:
### Changes
- Implemented a software rendering fallback for the background renderer.
  - Simply uses the "flat" background style colouring as it's nice and performant.
  - Should allow `SukiWindow` to not only be previewable in an IDE but should also let us run on Avalonia web in the future if that's desirable and things like VMs which don't have any hardware acceleration enabled.
  - Fixes: #288 
- Fixed a minor bug in the demo which prevented the title bar from toggling properly.

I've created an API for `SukiBackground` which allows you to force software rendering manually, however I've not exposed this through `SukiWindow` yet, if you think it's a good idea and might be useful I'm happy to add that before merging.

### DIFF
--- a/SukiUI.Demo/Controls/LoadingTest.cs
+++ b/SukiUI.Demo/Controls/LoadingTest.cs
@@ -63,6 +63,11 @@ namespace SukiUI.Demo.Controls
                 }
                 canvas.Restore();
             }
+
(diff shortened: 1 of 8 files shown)


kikipoulet:
very nice, I think it's not a bad idea to give the option through SukiWindow indeed 👍

PR #297, opened 2024-10-05
Switch old loading style with shader-based approach.

sirdoombox:
### Changes
- Replaced old Avalonia animation based loading indicator with a shader based one.
- Included two basic implementations in `Glow` and `Pellets` which aren't final and just show it working.
- Replaced all instances of the old loading setup with the new one, specifically in the buttons.

### Outstanding issues
- The current approach is to use existing uniforms to hack something together, this does unfortunately mean we've got a bunch of essentially duplicated shader files with just the final output colour changed.
- There is _some_ scope for me to change the API of `SukiEffect` to allow for custom uniforms but it'd end up being a pretty major change in order to support it, if this is what's desired I can go ahead with this as part of this PR. It shouldn't be a breaking change but it will take a bit more work to include support.
- The versions included aren't final and are just heavily adapted from things I found on ShaderToy, if you want to browse around to look for an implementation that's a bit more in-line with what you had in mind in terms of design then I recommend starting there.
- It looks pretty terrible on the buttons in light mode atm because I'm relying on the `iDark` uniform which isn't perfect. Though to be honest the light theme in general is a bit of a mess in places.


kikipoulet:
This is great, I wanted to avoid destroying the SukiEffect file myself with dirty hacks of mine with uniforms (which I did when I tested it actually) ..

Is there a way to have a Foreground property ? It would make it easier to style buttons loading and to use the resource colors of dark/light themes.

Don't worry about the shaders, I will modify it later to match with our design.

sirdoombox:
I can expand the API to allow custom uniforms for proper colours, it shouldn't be too huge a problem but it will take a little work to get it right.

Hopefully the examples working as-is should be a good first look at it running.

I'd like to see the buttons get changed alongside this change to the loading icon, it would be nice if the text scaled down to fit inside the button as-sized or just shifted to the right in some way, the resizing of the button for this feature is really awkward. I can look into it myself if it's not something you've got time for but I feel like this is the best time to fix this one.

sirdoombox:
This is mostly in a good place now, functioning as intended.

I also took the opportunity to grab some low hanging fruit in terms of performance, now all our own runtime effect uniforms use pre-allocated arrays that are just updated as needed, this saves a nice chunk of per-frame alloc and GC work (I should've done it this way to begin with).

Issue #167, opened 2024-04-12
High CPU usage issues


Cardroid:
**Describe the bug**
I'm currently working on a program that draws audio plots.
I'm aware of the high CPU utilization because of the work of drawing cool animations.

However, the program seems to be constantly drawing new frames on a still screen (even when there is no user input).
CPU seems to be being wasted unnecessarily due to continuous frame refreshes.
How can we improve it?

**To Reproduce**

1. Apply the Suki library and run the program.

**Expected behavior**
CPU utilization should be close to 0% on a still screen without events

**Screenshots**
without SukiUI
![low cpu](https://github.com/kikipoulet/SukiUI/assets/25889754/99b5d982-a878-4d78-a5dc-1bd7ca114646)
CPU usage: 0% ~ 1%

with SukiUI
![high cpu](https://github.com/kikipoulet/SukiUI/assets/25889754/7f8e6284-b14d-4760-9768-24be97473850)
CPU usage: 5% ~ 6%

** Environment
- OS: Windows 11
- Avalonia-Version: 11.0.10

The plotting library used [ScottPlot](https://github.com/ScottPlot/ScottPlot)

kikipoulet:
Hi,

Unfortunately I noticed this behavior a few times ago. The problem is that it does not seem to be one thing in particular that trigger this, but the overall complexity of the logical tree. It's not just removing the dynamic background, or the dialog grid, that can solve this, everything that we use "add" a little complexity and it ends at 4-5%, even on my machine.

Well, it was my observation at least. But performance is a feature, and I kept that problem in my mind, it does not seem normal anyway. I think you maybe found more with the rendering observations, so if I summary well, SukiUI suddenly renders insane amount of frame there and there when normally it would cap at 30 fps ?

Moreover, can you try to use SukiUI theme without using SukiWindow to see what it produces ? just to see if it changes something, I suspect that the SukiWindow is the source of 90% of the performance problem.  

It would be great if we find a way to fix this actually 👍

Cardroid:
Thank you for your answer. 👍

I also want to continue using this wonderful library.
After the development of the core BM, I will investigate it.

This is just my guess... I think it's related to background animation. (when I see a lot of GC calls at the beginning of the program running)

kikipoulet:
Thanks for your kind words.

I don't think it's related to the background animation because it's just at startup, and removing it doesn't change anything, but maybe the fps mystery is related

kikipoulet:
[...]

PR #310, opened 2024-10-16
Rewrite internals of `EffectDrawBase`

sirdoombox:
### Changes
- Internally, `EffectDrawBase` now uses `CompositionCustomVisualHandler`.
  - Whenever animations are disabled, GPU & CPU usage are significantly lower (near 0%) - Related Issue: #167 
- Fix `SukiTransitioningContentControl`.
  - Transition animations now play correctly.
  - Backbuffer is cleared out when it's done playing it's animation and assigning `null` to the content is now valid which should make detaching content from the visual tree work as intended.

### Outstanding Issues
- The `Loading` control - even when invalidating only the rect for it - still has very high GPU utilisation and I'm not quite sure why.
- The `ShaderToyRenderer` in the demo also suffers from this same issue - though it seems to use less GPU for reasons beyond my understanding.
- We don't really have a docs page covering `SukiWindow` at the moment.

I wanted to ideally have at least one or two other people test this to make sure the GPU/CPU issues are in fact under control and everything renders correctly, before this gets merged as it's a pretty major internal change.

### DIFF
--- a/SukiUI/Utilities/Effects/EffectBackgroundDraw.cs
+++ b/SukiUI/Utilities/Effects/EffectBackgroundDraw.cs
@@ -14,14 +14,17 @@ namespace SukiUI.Utilities.Effects
         internal bool TransitionsEnabled { get; set; }
         internal double TransitionTime { get; set; }
 
-        private static readonly Stopwatch TransitionTick = Stopwatch.StartNew();
-
-        private static float TransitionSeconds => (float)TransitionTick.Elapsed.TotalSeconds;
+        private float TransitionSeconds => (float)CompositionNow.TotalSeconds;
 
         private SukiEffect? _oldEffect;
         private float _transitionStartTime;
         private float _transitionEndTime;
 
+        public EffectBackgroundDraw() : base(true)
+        {
+            
+        }
+        
         protected override void EffectChanged(SukiEffect? oldValue, SukiEffect? newValue)
         {
             if (!TransitionsEnabled) return;
--- a/SukiUI/Utilities/Effects/EffectBackgroundDraw.cs
+++ b/SukiUI/Utilities/Effects/EffectBackgroundDraw.cs
@@ -20,7 +20,7 @@ namespace SukiUI.Utilities.Effects
         private float _transitionStartTime;
         private float _transitionEndTime;
 
-        public EffectBackgroundDraw() : base(true)
+        public EffectBackgroundDraw() : base(false)
         {
             
         }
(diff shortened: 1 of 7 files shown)


kikipoulet:
No problem on my side, behave like expected with great performance improvements

sirdoombox:
I annoyingly discovered a bug with `SukiTransitioningContentControl` whilst I was testing this out in another app, the first time content was set `IsHitTestVisible` was being set to false and I'm not fully sure why, struggling to actually get it to enable correctly. 

Whilst trying to fix this bug I also discovered that the core functionality was completely disabled and broken in #259 for some reason. I think the bug this was supposed to fix needs a solution, hopefully I can come up with one that retains the functionality of the transitioning content control.

sirdoombox:
I'm not 100% certain I've fixed the issue that #259 was supposed to fix, but it _should_ do the job, now the back buffer is cleared out and you can assign `null` to the `Content `property which should cause the content to be cleared out correctly. It also re-enables the transition animations and such.

sirdoombox:
Could you confirm that #259 isn't regressed with these changes @AuroraZiling? If the crash is back, please send me a repro and I'll try and implement a better solution.

AuroraZiling (approved):

SukiUI/Utilities/Effects/EffectDrawBase.cs
```diff
@@ -51,30 +53,54 @@ public bool AnimationEnabled
...
+            else if (message is SukiEffect effect)
+            {
+                Effect = effect;
+            }
+        }
+
+        public override void OnAnimationFrameUpdate()
+        {
+            if (!AnimationEnabled) return;
+            Invalidate(GetRenderBounds());
```
maxkatz6, line 99: Unless you need to invalidate only small portion of the visual, you can just call `Invalidate()` method. It will reduce some overhead on calculating dirty rects per visual.

SukiUI/Utilities/Effects/EffectBackgroundDraw.cs
```diff
@@ -33,10 +31,16 @@ protected override void EffectChanged(SukiEffect? oldValue, SukiEffect? newValue
             _transitionEndTime = TransitionSeconds + (float)Math.Max(0, TransitionTime);
```
maxkatz6, line 31: You can replace `TransitionSeconds` with `CompositionCustomVisualHandler.CompositionNow` timespan property, which is syncrhonised with render timer. Probably won't improve anything perf-wise though.

sirdoombox:
Thanks for the suggestions @maxkatz6, PR is ready to go.

I won't merge it myself just so one final sanity check gets done before it gets merged.

kikipoulet:
> Thanks for the suggestions @maxkatz6, PR is ready to go.
> 
> I won't merge it myself just so one final sanity check gets done before it gets merged.

I tested it, LGTM

PR #327, opened 2024-11-08
Bugfix for #324 

Insire:
- Fixed PropertyGrid not opening complex child viewmodel in new dialog window
- Added SukiHost support to PropertyGrid (back)
- Adresses #324


sirdoombox:
LGTM.

PR #330, opened 2024-11-11
Simplify `Flat` shader. Include `Background` docks page.

sirdoombox:
### Changes
- Simplifed the `Flat` shader, there was no need for the `iDark` branch nor the `* 1` assignment. The underlying `iBase` colour handles everything for us.
  - This should actually marginally improve performance as GPUs hate branching.

### Docs
- Included a `Background` page in theming to walk people through the basics of changing the background style and how custom shaders work.
  - Also includes a warning about animations.


kikipoulet:
Awesome ! 😳

PR #331, opened 2024-11-19
Support for RTL Layout

dme-compunet:
SukiUI is one of the beautiful UIs I've seen, so it would be great if it supported RTL layout.

Note: In the current avalonia version of the project, there are some rendering issues in RTL mode that have been fixed in version 11.2.0. Please check this PR with the latest avalonia release.


sirdoombox:
RTL is not something I have a huge deal of experience with, but having checked out the commit locally, everything seems to be working as intended.

This will be a fairly narrowly used feature, but supporting arabic/hebrew/etc. is probably a good idea, the fact this also toggles quite nicely and everything updates correctly is a big plus for runtime localisation.

LGTM, but I'd want @kikipoulet to weigh in on a big feature addition instead of just merging it myself.

kikipoulet:
Looks good to me too, that's great 👍

PR #342, opened 2024-11-30
[Housekeeping] for project files and package generation

Insire:
### Relevant changes
- added Directory.Build.props, Directory.Packages.props to centrally manage nuget package versions and global project settings (speeds up managing nuget packages via IDE dramatically)
- added [*.{csproj,props}] to editorconfig, to enforce uniform formatting for props and csproj files
- auto formatted csproj and props files according to new editorconfig settings

### Breaking Changes
- use netstandard2.0 for SukiUI.Dock
- added tags to SukiUI.Dock

### Misc
- fixed ambiguous reference in ToastsViewModel
- added most recent .NET 8 version of System.Text.Json to Demo app, to fix a transitive package vulnerability

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,9 +1,15 @@
 root = true
 
+[*]
 tab_width = 4
 indent_size = 4
 end_of_line = crlf
 indent_style = space
+
+[*.{csproj,props}]
+indent_style = space
+tab_width = 2
+indent_size = 2
 
 [*.{cs,vb}]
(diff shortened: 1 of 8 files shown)


Insire:
Not sure why SukiUI.Dock was using .NET 8.0, when SukiUI itself was .netstandard2.0. So i upgraded that. If there is a specific reason why it should remain on .NET 8.0, i will happily revert that change.

PR #351, opened 2024-12-09
UI Button

dme-compunet:
- Change the cursor to Hand (when pointer over)
- Add icon button class

### DIFF
--- a/SukiUI.Demo/Features/ControlsLibrary/ButtonsView.axaml
+++ b/SukiUI.Demo/Features/ControlsLibrary/ButtonsView.axaml
@@ -17,7 +17,22 @@
             <Setter Property="Command" Value="{Binding ButtonClickedCommand}" />
             <Setter Property="IsEnabled" Value="{Binding IsEnabled}" />
         </Style>
+        <Style Selector="Button.Icon">
+            <Setter Property="Content">
+                <Template>
+                    <PathIcon Margin="0 3 0 0" 
+                              Data="{StaticResource HeartIcon}"
+                              Foreground="{Binding Foreground, RelativeSource={RelativeSource FindAncestor, AncestorType=Button}}"/>
+                </Template>
+            </Setter>
+        </Style>
+
     </UserControl.Styles>
+    <UserControl.Resources>
+        <StreamGeometry x:Key="HeartIcon">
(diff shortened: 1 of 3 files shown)


deepea:
This is just my opinion but, while I appreciate the effort in the PR, I don't agree with the hand cursor changes:
- Non-standard. On Windows, at least, this is not expected behavior
- Arbitrary. e.g. Why Button but not ToggleButton? Why RadioButton but not CheckBox?
- It affects other controls derive from them. e.g. GigaChip

dme-compunet:
It's interesting to hear what the general opinion is here, If the majority of users don't agree with this, I'll just return this pledge, leaving only the icon button.

sirdoombox:
> This is just my opinion but, while I appreciate the effort in the PR, I don't agree with the hand cursor changes:
> 
>     * Non-standard. On Windows, at least, this is not expected behavior
> 
>     * Arbitrary. e.g. Why Button but not ToggleButton? Why RadioButton but not CheckBox?
> 
>     * It affects other controls derive from them. e.g. GigaChip

I'm in agreement here, I think this is a nice idea but isn't really suitable.

sirdoombox:
Just a couple small notes on the icon button: if that 3px top margin is necessary every time you use pathicon it might be an idea to include that margin in the icon button style with a selector for pathicon; `flat` and `flat rounded` seem to be functionally identical in this case so it's probably not worth demoing; the accent colour variations aren't in the demo either which might be an idea; finally we already have a load of standard icons in SukiUI you can use for demoing the icon button to save some lines of code e.g.
```xaml
<PathIcon Data="{x:Static Icons.Star}" />
```
Other than that I really like the idea of the icon button, looks neat and tidy and fits in nicely with the design.

dme-compunet:
I don't understand why, but in accent variations the outlined button doesn't change the icon foreground when pointer over

sirdoombox:
I also see now that the standard color buttons don't seem to have working `:pointerover` effects, so there's no visual feedback when you mouse over them...

Hmm I think buttons in general are going to need a bit of a polish pass because it's a bit of an inconsistent mess right now.

I don't think I'll have time to do it personally until at least the weekend.

kikipoulet:
> Yes there's probably a special case there somewhere in the file for handling that effect...
> 
> I also see now that the standard color buttons don't seem to have working `:pointerover` effects, so there's no visual feedback when you mouse over them...
> 
> Hmm I think buttons in general are going to need a bit of a polish pass because it's a bit of an inconsistent mess right now.
> 
> I don't think I'll have time to do it personally until at least the weekend.

We're talking about the "Flat" button that is the only class that is not animated on hover right ? Maybe a slight background change to SukiPrimaryColor90 on pointerover would be just right ?

sirdoombox:
> We're talking about the "Flat" button that is the only class that is not animated on hover right ? Maybe a slight background change to SukiPrimaryColor90 on pointerover would be just right ?

Yep, I'm not sure what it is for `Accent` but the standard coloured `Flat` buttons should match that.

Could probably accept this PR as is and fix up the bits and pieces if need be, the only outstanding issue with the PR itself is the accent button's content isn't having it's colour set correctly.

PR #354, opened 2024-12-11
Change standard flat button style to match accent.

sirdoombox:
### Changes
- Adjusted standard `Flat` style button to match the `Accent` one.
- Fixed weird naming of a converter.

I'm still not that happy with the implementation of the busy button spinner suddenly taking up unreserved space, it's especially weird looking in the new icon button when it just pushes the icon out.

I'm not sure there's a way to implement it in a tidy way though, so I guess it is what it is.

### DIFF
--- a/SukiUI/Controls/BusyArea.axaml
+++ b/SukiUI/Controls/BusyArea.axaml
@@ -62,7 +62,7 @@
                             
                             <ContentPresenter HorizontalAlignment="Center"
                                               VerticalAlignment="Center"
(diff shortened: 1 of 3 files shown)


dme-compunet:
@sirdoombox How about it be the other way around, i.e.:
- static: `SukiPrimaryColor`
- pointerover: `SukiPrimaryColor75`
- pressed: `SukiPrimaryColor50`

(the `:pressed` state can also remain on `SukiPrimaryColor75`)

PR #365, opened 2024-12-17
SideMenuItem: Set FontSize to a bindable value

SwaggyMacro:
The default FontSize of the Header is too small on WindowsPad, and it cannot be customized.

Default FontSize(14):
159c11215fb5f4aa39ebce89c2cc548

FontSize(20):
095aad6cd2ce870e518ecbf439fec58


kikipoulet:
Hi, 

We have no specific plan on supporting touch devices, it is out of our scope. However, I have nothing against merging PR that add customization where it should be, just be sure that the default appareance stay still.

Can you modify your PR to let the FontSize 14 by default ?

SwaggyMacro:
> Hi,
> 
> We have no specific plan on supporting touch devices, it is out of our scope. However, I have nothing against merging PR that add customization where it should be, just be sure that the default appareance stay still.
> 
> Can you modify your PR to let the FontSize 14 by default ?

https://github.com/kikipoulet/SukiUI/pull/365 the new one

Issue #363, opened 2024-12-13
[Bug] DataGrid styles inconsistencies 


sn4k3:
### Check the following items

- [X] I have looked up relevant Issue

### Description of the issue

Found some inconsistencies in styles of DataGrid

1. Lack of styles for groups (`DataGridCollectionView`)
2. Size of cells are incorrect (Overflowing) causing bad alignment and text out of cell limit 

![image](https://github.com/user-attachments/assets/a6308378-b1ae-4224-b9d0-50837d67bf72)

### Package Version

6.0.0-beta6

### Environment

Windows 11

### Expected Behavior

1. Group expander to have matching style
2. Cells to have proper bounds

### Reproduction

```C#
public partial class TestUserControl : UserControl
{
    public class TestClass
    {
        public int Index { get; set; }
        public string Name { get; set; }
        public string Description { get; set; }
        public string Group { get; set; }
    }

    public ObservableCollection<TestClass> TestCollection { get; } =
    [
        new() { Index = 1, Name = "Test 1", Description = "Description 1", Group = "Group 1"},
        new() { Index = 2, Name = "Test 2", Description = "Description 2", Group = "Group 1" },
        new() { Index = 3, Name = "Test 3", Description = "Description 3", Group = "Group 2" },
        new() { Index = 4, Name = "Test 4", Description = "Description 4", Group = "Group 2" },
    ];


    public TestUserControl()
    {
        InitializeComponent();
        DataContext = this;
    }

    protected override void OnLoaded(RoutedEventArgs e)
    {
        base.OnLoaded(e);
        var groupView = new DataGridCollectionView(TestCollection);
        groupView.GroupDescriptions.Add(new DataGridPathGroupDescription("Group"));
        MyGrid.ItemsSource = groupView;
    }
}
```

```xml
<UserControl xmlns="https://github.com/avaloniaui"
             xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
             xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
             xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
             xmlns:views="clr-namespace:NetSonar.Avalonia.Views"
             xmlns:suki="https://github.com/kikipoulet/SukiUI"
             mc:Ignorable="d" d:DesignWidth="800" d:DesignHeight="450"
             x:Class="NetSonar.Avalonia.Views.TestUserControl"
             x:DataType="views:TestUserControl">
[...]

PR #370, opened 2024-12-28
Fix DataGrid GroupRow styles

sn4k3:
This PR aim to fix the DataGrid GroupRow styles as explaned in #363

(Help wanted) It still has some annoyances I can't get to work:
- When click on expander arrow it will change arrow state but does not expand nor hide
- When click on header it will expand and hide but not change arrow state
- Missing Suki animations for expander

https://github.com/user-attachments/assets/564bf01d-82b4-4d28-b140-89d64aa2bb69

As we don't include DataGrid theme from Avalonia there are many things missing, as for example, we can't aim `DataGridRow` type for inline styling of row because it does not exists as ControlTheme: `<ControlTheme x:Key="{x:Type DataGridRow}" TargetType="DataGridRow">`.
A better approach would be include default DataGrid theme and override the key parts on Suki, or copy over from Avalonia and change every element to match Suki.
The [Fluent theme](https://github.com/AvaloniaUI/Avalonia/blob/master/src/Avalonia.Controls.DataGrid/Themes/Fluent.xaml) of DataGrid already include many ResourceDictionary that we can tune, however some key parts are still missing.

Sample effect of lacking stock ControlThemes:
image

### DIFF
--- a/SukiUI.Demo/Features/ControlsLibrary/CollectionsView.axaml
+++ b/SukiUI.Demo/Features/ControlsLibrary/CollectionsView.axaml
@@ -72,15 +72,16 @@
                                           MaxHeight="200"
                                           CanUserResizeColumns="{Binding IsDataGridColumnsResizable}"
                                           ItemsSource="{Binding DataGridContent}">
+                                    <DataGrid.Styles>
+                                        <Style Selector="DataGridRowGroupHeader">
+                                            <Setter Property="IsItemCountVisible" Value="True" />
+                                        </Style>
+                                    </DataGrid.Styles>
                                     <DataGrid.Columns>
-                                        <DataGridTextColumn Binding="{Binding StringColumn}" Header="String" />
-                                        <DataGridTextColumn Binding="{Binding IntColumn}"
-                                                            
-                                                            Header="Int" />
-                                        <DataGridCheckBoxColumn Binding="{Binding BoolColumn}"
-                                                               
-                                                                Header="Bool" />
-                                        
+                                        <DataGridTextColumn Binding="{Binding StringColumn, DataType=controlsLibrary:DataGridContentViewModel}" Header="String" />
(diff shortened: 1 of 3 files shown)


kikipoulet:
I'm not in a hurry to take a look at this as I'm on vacation but we will definitely try to support DataGrid correctly.

You seem to have a deep usage of DataGrid control, I'm really wondering by curiosity what is your use-case and what you're trying to do with this control.

kikipoulet:
I did a first rough fix. Modifications will be needed to polish style at least

sn4k3:
> You seem to have a deep usage of DataGrid control, I'm really wondering by curiosity what is your use-case and what you're trying to do with this control.

Nothing fancy... I have a ping collection and I like to set row background to Red when a ping fails otherwise stays normal, this is impossible due lack of styles (Also not possible with code behind). I try to set colors per item status which also impossible due the lack of ControlTheme as shown on image.
Using groups to stack per user defined groups.

As the theme is missing some key components from base makes it poor for customization.

kikipoulet:
Well, isn't it about making a nice fancy datagrid design ? If it's an open-source project I would be curious to look at the repo.

I will look for a row color solution, don't forget to try the last quick fix I did to see if it's what you wanted

sn4k3:
It will be open-source but normally I only push repo online on first release (Which is a bit far given my concept).
But if you wish I can send you a source copy.
It's easy to replicated by this code:

```xml
<DataGrid.Styles>
    <Style Selector="DataGridRow">
        <Setter Property="Background" Value="DarkRed" />
    </Style>
</DataGrid.Styles>
```

To change background based on item status we can use:

```xml
<DataGrid.RowTheme>
    <ControlTheme
        x:DataType="network:PingableService"
        BasedOn="{StaticResource {x:Type DataGridRow}}"
        TargetType="DataGridRow">
        <Setter Property="Background" Value="{Binding WasLastPingSucceeded, Converter={StaticResource BoolErrorGridRowBackground}}" />
    </ControlTheme>
</DataGrid.RowTheme>
```

However, the last will not compile due lack of DataGridRow ControlTheme.

PR #380, opened 2025-01-13
Update SukiWindow.axaml.cs

PulpFiction98:
Fix: Unable to change window size with mouse after switching between full screen and normal mode


niuzhaosen:
The size of the title bar and mouse drag memory should be controlled by the user themselves

Issue #368, opened 2024-12-20
[Bug] Unhandled Exception with SukiUI Dock Control: "AttachedToLogicalTreeCore called for 'Document' but control has no logical parent"


envy666:
### Check the following items

- [X] I have looked up relevant Issue

### Description of the issue

Dear SUKIUI Team,

I am writing to report an issue encountered while using the SukiUI Dock control in my project. The problem I faced is an unhandled exception which causes the application to crash. The error message reads:

"AttachedToLogicalTreeCore called for 'Document' but control has no logical parent"

Initially, I suspected that the issue might be due to a mistake in my implementation, so I carefully compared my usage with the official DEMO provided by SukiUI. However, during this verification process, I managed to reproduce the same issue within the official DEMO itself.

Steps to Reproduce:
Drag the Program.cs form to float it outside of its original docked position.
Drop the form back into the Document area.
Switch the side menu to any other menu item.
Switch back to the dock menu containing the previously floated and re-docked document.
Upon performing these steps, the application crashes at SukiTransitioningContentControl.SecondBuffer.SetValue.

### Package Version

6.0.0-beta8

### Environment

win10

### Expected Behavior

fixed bug

### Reproduction

Steps to Reproduce:
Drag the Program.cs form to float it outside of its original docked position.
Drop the form back into the Document area.
Switch the side menu to any other menu item.
Switch back to the dock menu containing the previously floated and re-docked document.
Upon performing these steps, the application crashes at SukiTransitioningContentControl.SecondBuffer.SetValue.

AuroraZiling:
I can't reproduce this issue

AuroraZiling:
I can't reproduce this issue in 6.0.0-beta8 and latest dev in demo app (win11 24h2)

envy666:
https://github.com/user-attachments/assets/6f74306b-273d-446f-b33c-e9a038197408

Issue #382, opened 2025-01-14
Got in the Demo [Bug] System.InvalidOperationException: AttachedToLogicalTreeCore called for 'Tool' but control has no logical parent.


orangebobolink:
### Check the following items

- [X] I have looked up relevant Issue

### Description of the issue

Unhandled exception. System.InvalidOperationException: AttachedToLogicalTreeCore called for 'Tool' but control has no logical parent.
   at Avalonia.StyledElement.OnAttachedToLogicalTreeCore(LogicalTreeAttachmentEventArgs e)
   at Avalonia.StyledElement.OnAttachedToLogicalTreeCore(LogicalTreeAttachmentEventArgs e)
   at Avalonia.StyledElement.OnAttachedToLogicalTreeCore(LogicalTreeAttachmentEventArgs e)
   at Avalonia.StyledElement.OnAttachedToLogicalTreeCore(LogicalTreeAttachmentEventArgs e)
   at Avalonia.StyledElement.Avalonia.LogicalTree.ILogical.NotifyAttachedToLogicalTree(LogicalTreeAttachmentEventArgs e)
   at Avalonia.Controls.Primitives.TemplatedControl.OnAttachedToLogicalTree(LogicalTreeAttachmentEventArgs e)
   at Avalonia.StyledElement.OnAttachedToLogicalTreeCore(LogicalTreeAttachmentEventArgs e)
   at Avalonia.StyledElement.OnAttachedToLogicalTreeCore(LogicalTreeAttachmentEventArgs e)
   at Avalonia.StyledElement.OnAttachedToLogicalTreeCore(LogicalTreeAttachmentEventArgs e)
   at Avalonia.StyledElement.Avalonia.LogicalTree.ILogical.NotifyAttachedToLogicalTree(LogicalTreeAttachmentEventArgs e)
   at Avalonia.Controls.Primitives.TemplatedControl.OnAttachedToLogicalTree(LogicalTreeAttachmentEventArgs e)
   at Avalonia.StyledElement.OnAttachedToLogicalTreeCore(LogicalTreeAttachmentEventArgs e)
   at Avalonia.StyledElement.OnAttachedToLogicalTreeCore(LogicalTreeAttachmentEventArgs e)
   at Avalonia.StyledElement.OnAttachedToLogicalTreeCore(LogicalTreeAttachmentEventArgs e)
   at Avalonia.StyledElement.OnAttachedToLogicalTreeCore(LogicalTreeAttachmentEventArgs e)
[...]

PR #386, opened 2025-01-21
Add Dock Mvvm sample to SukiUI.Demo

orangebobolink:
It can fix issue with [System.InvalidOperationException](https://github.com/kikipoulet/SukiUI/issues/382)


orangebobolink:
It also can fix #368

kikipoulet:
Great !

PR #396, opened 2025-01-27
Add save as and open example for SukiUi.Dock to SukiUi.Demo

orangebobolink:
This example was added for Mvvm
There is alreade an example for Xaml [here](https://github.com/wieslawsoltes/Dock/tree/master/samples/DockXamlSample)


kikipoulet:
Nice to see some Dock support

PR #402, opened 2025-02-03
Add basic localization support

### LINKED ISSUE #372: [Feat] i18n support for some build-in text

- [X] I have searched for duplicate issues.

Some built-in controls in Avalonia contain text information (such as the default right-click menu of TextBox).

And I find that all of the text is hard-coded in UserControl. So it's hard to make contributions to add language.
We needs to investigate which solution can be apply.

Do you have any ideas on it?

### Package Version

6.0.0

realybin:
Fixes #372 Add basic localization support for some build-in control textual content
Specific implementation and reference from another great control library [Semi.Avalonia](https://github.com/irihitech/Semi.Avalonia) with MIT license


realybin:
Thank you!

Issue #418, opened 2025-03-04
[Bug] Fullscreen toogle breaks CanMaximize, CanResize and CanMove


sn4k3:
### Check the following items

- [x] I have looked up relevant Issue

### Description of the issue

Toogle fullscreen on and off, permanently disable `CanMaximize`, `CanResize` and `CanMove`.
Problem: 
https://github.com/kikipoulet/SukiUI/blob/6928c6b0404c027c1db55801b1c816cf0f134cc8/SukiUI/Controls/SukiWindow.axaml.cs#L354-L362

Also, title bar is not hidden by default, which should be...

### Package Version

6.0

### Environment

Windows 11

### Expected Behavior

After reverting to any other than Fullscreen to restore `CanMaximize`, `CanResize` and `CanMove` to the old properties.

### Reproduction

```C#
WindowState = WindowState.FullScreen;
WindowState = WindowState.Normal;
```

### Additional Information

Can be fixed by implementing a cache of previous values and restaure after exit fullscreen.

Waheal:
This problem has been fixed, but the author forgot to close this issue. Now you can close it.

Issue #387, opened 2025-01-22
[Bug] After maximizing on MacOS, the Window background will appear around the form(?)


imJack6:
### Check the following items

- [x] I have looked up relevant Issue

### Description of the issue

There is no such problem in the normal non-maximized state.
![Image](https://github.com/user-attachments/assets/fba4cea6-6323-44bb-b171-d005a1561cd4)
![Image](https://github.com/user-attachments/assets/813e31d0-e86b-470e-b03f-28baafeac36a)

When I set the TransparencyLevelHint="AcrylicBlur" of SukiWindow, the edge will also become transparent or blurred. When I delete this attribute, it will become dark gray, and when I adjust the system theme to Light, the edge will also become white synchronously.

### Package Version

6.0.0-rc

### Environment

MacOS 15.2

### Expected Behavior

Maximize the perfect display.

### Reproduction

On MacOS use SukiWindow and then maximize the window.

Waheal:
This problem has been fixed, but the author forgot to close this issue. Now you can close it.

PR #421, opened 2025-03-05
Update for SukiWindow

Waheal:
1.Added ResizeGrip for SukiWindow on **Linux** to let window can resize.

2.Fixed the bug where the false property of CanMove failed to take effect.

3.**Only** set margin to 7 when WindowState is Maximized **on Windows platform**.

4.Added temporary variables to store CanMaximize and CanMove states before entering FullScreen mode, and restore them when WindowState returns to Normal. (It is not necessary to change the CanResize property in FullScreen State because it seems that users cannot change the window size in this state)

5.Made RootCornerRadius independently configurable (recommended to set only for Linux systems. It will auto-set to 10 on Linux if not explicitly configured).

### DIFF
--- a/SukiUI/Controls/SukiWindow.axaml
+++ b/SukiUI/Controls/SukiWindow.axaml
@@ -4,10 +4,8 @@
                     xmlns:suki="clr-namespace:SukiUI.Controls">
     <ControlTheme x:Key="SukiWindowTheme" TargetType="suki:SukiWindow">
         <Setter Property="Margin" Value="0" />
-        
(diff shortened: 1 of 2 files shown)


kikipoulet:
Sorry for late merge, I've been quite busy recently.

Super fix, thank you very much for your contribution 👍

Waheal:
Thank you for merging the PR! I completely understand your busy - no worries about the timing at all. 😊

Just wanted to mention this fix should resolve the issues:
- #418
- #387

Feel free to close those issues if the solutions work as expected. Happy to contribute and looking forward to future collaborations!

PR #433, opened 2025-03-21
Improvements to InfoBar

sn4k3:
- Bind `HorizontalContentAlignment` and `VerticalContentAlignment` to allow aligns
- Add `TitleOrientation` to be able to stack title vertically or horizontally
- Add `MessageTextAlignment` to allow text alignment
- Add `IconSize` to allow change icon size
- Add `IsTextSelectable` to allow select text
- Change grid column width to Auto instead of fixed values
- Fix ghost icon when null (Hide in a null situation)

image


kikipoulet:
🙏🙏🙏

PR #438, opened 2025-03-25
[Feat] Improve window system buttons looks

dme-compunet:


sirdoombox:
LGTM.

The only truly major change is the button cursor style, and as far as I know classic cursor on buttons is the standard behaviour in most UI frameworks now, and is the default in Avalonia's fluent + simple themes.

Issue #441, opened 2025-03-27
[Bug] Dialogs with ViewModel don't show action buttons added in builder


susloparovdenis:
### Check the following items

- [x] I have looked up relevant Issue

### Description of the issue

When run dialogManager.CreateDialog() with ViewModel
```csharp
  dialogManager.CreateDialog()
            .WithViewModel(dialog => new VmDialogViewModel(dialog))
            .WithActionButton("Close " + SelectedType.ToString(), _ => { }, true, "Flat")
            .TryShow();
```
no action buttons shown

### Package Version

6.0.1

### Environment

Windows 11

### Expected Behavior

buttons are shown

### Reproduction

```csharp
  dialogManager.CreateDialog()
            .WithViewModel(dialog => new VmDialogViewModel(dialog))
            .WithActionButton("Close " + SelectedType.ToString(), _ => { }, true, "Flat")
            .TryShow();
```

PR #442, opened 2025-03-27
Dialog fixes and improvements.

sirdoombox:
### Changes
- Now actually returns dialogs to the pool and reuses them.
  - It statically creates 2 at initialisation, without this we'd need to synchronise the animation system up with returning it to the pool, and I'd rather not do that.
- Dialogs can optionally use a ViewModel as content now by setting the `WithViewModel` optional parameter to false - Fixes: #441
- Tidy up the formatting a bit in SukiDialog.axaml


sirdoombox:
No breaking changes here, fairly minor internal changes only so I'll just merge this myself.

PR #446, opened 2025-03-28
Add nl-NL (Dutch) locale

RivenSkaye:
My use-case has me requiring Dutch, so I might as well upstream it right?


RivenSkaye:
Still a work in progress by the way - making sure I have all references right before committing those files (e.g. the Compile directive in the csproj was still missing)

kikipoulet:
As a walloon I studied dutch during 10 years. 

I don't remember even one word, 10/10 langage

 PR approved

PR #451, opened 2025-03-29
Add window max size ratio related to screen properties and extra fixes

sn4k3:
- Add `MaxWidthScreenRatio` and `MaxHeightScreenRatio` properties
  - If value is > 0, it will constrain the MaxWidth and/or the MaxHeight to fit inside the host screen, allowing tool windows (CanResize = False and SizeToContent=Auto) live happy inside a sane space, without this a extra effort is required to make sure window never grow beside screen size which is hard to know in every case. As I heavy use tool windows in my apps I'm adding this PR to bring same function to SukiUI. This allow create nice and sane dialog windows.
  - It will update max size when scaling and screen change by user, but if window is moved by other means it will not until user move it with title bar inteaction, I had this working but avalonia has a "[bug](https://github.com/AvaloniaUI/Avalonia/issues/18552)" that can cause more bad than good
  - Another flaw is if change to fullscreen or maximize, the max size is updating correctly but it does after the event so fullscreen or maximize will be limited to old max limits. This can be fixed by user or if we have a way to access WindowState before change event.
- Add `LogoDoubleTapClosesWindow` Property to mimic windows native double click on logo, which closes the window. Default: false
- Add null safe check for `IsPointerOver` (Should never happen but will clear the nullable warning)
- Add pt_PT locale
- Do not cache nor change CanMove, CanResize and CanMaximize
  - We should not change user settings in most cases, and the implementation still was flaws like Maximize still CanResize=true when it can't really resize, so it doesn't make much sense to cover some sides while not on others. The restrictions are now directly applied on code and xaml styles. 
- Dispose all events in a case of theme reapply
- Fix EnableWindowsSnapLayout by `if (!maximize.IsVisible) return 0;`, before, it was still trigger maximize when CanMaximize is False and we click in it ghost region.
- Remove `Basic Rounded` extra classes from new buttons as they are no longer required due a recent PR
- Add `WindowExtensions.cs` with two methods

To test this PR:
1. Open demo app
2. Dialogs
3. Open Tool Window demo

https://github.com/user-attachments/assets/3cff6cbe-3b93-46c6-80ab-b450b1b96595


kikipoulet:
Looks great 👍

Issue #394, opened 2025-01-26
[Feat] Toast option to not dismiss while mouse hover


sn4k3:
### Use the search function to see if this feature has already been suggested.

- [x] I have searched for duplicate issues.

### Description of new feature

Toast popups can be tricky to read when they have a dismiss delay, also for who is a slow reader and/or not native to language would benefit from a paused dismiss when hover the mouse over the toast to be able to get all the information. This is also useful when having actions on toast but not want them to stay there until manual dismiss.

I suggest a bool variable on `After` method to implement this, eg: `SukiToastBuilder After(this SukiToastBuilder.DismissToast dismiss, TimeSpan after, bool pauseWhileHover = true)` which when true and user hover the toast it just stops the timer (eg. DispatcherTimer). When it lost hover timer is reset.
Also, I think `true` should be the default, as this hover to pause dismiss is what modern systems uses with time notifications.

### Package Version

6.0 RC

sn4k3:
Implemented.

PR #455, opened 2025-04-02
SukiToast: Rework dismiss by time and allow to interrupt when hover pointer

sn4k3:
- Timed Dismiss is now controlled by SukiToast and not an outsider Task (`bool CanDismissByTime` and `TimeSpan DismissTimeout`)
- Allow to interrupt timed dismiss when hover pointer (`bool InterruptDismissTimerWhileHover`) (Implements #394)
- Add a progress as visual hint for the timed dismiss
- Get rid of the multiple DismissAction methods and merge into one with additional enum that tell the source of the dismiss (`Action<ISukiToast, SukiToastDismissSource>?`)

https://github.com/user-attachments/assets/486674a0-9105-4804-ac39-0dd61f6a1e99


Insire:
I'm not a fan of creating a timer for each toast and having it run on the ui-thread. I believe performance is a feature and with your changes, you are needlessly giving some of that up.

sn4k3:
> I'm not a fan of creating a timer for each toast and having it run on the ui-thread. I believe performance is a feature and with your changes, you are needlessly giving some of that up.

Toasts are re-used and timers only run if they had to. 
By using timed dismiss with old code a Task is created, this is "no different" by using a Timer which also creates a Task and fires it on UI thread. However that results in having larger object due extra DispatcherTimer instance.
If you never use the dismiss by time you will have unnecessary larger SukiToast object
The biggest performance gap on this PR is the progress value updater which actualy does at 50ms to present a fluild bar.

I can propose:

- Use `Lazy<DispatcherTimer>` so it will never instantiate the timer if you never use such feature, can also null them on Unloaded if required...
-  Add a property to whatever use the progress bar indicator, if false it will release the extra effort
- Create a single Timer on Manager (maybe with 500ms interval) that active prompt all active Toasts, to dismiss them when the time is right and update the progress value as well.

Insire:
The old version creates a task yes, but that task by default runs on the threadpool, not the ui-thread. Only the continuation is being marshalled back to the ui-thread.

I think having a single timer (not a dispatcher timer) that checks which toasts should be automatically dismissed is a solid approach. When a toast is to be dismissed, that logic should be marshalled to the ui-thread via dispatcher. Ideally, if there are a lot of Toasts, that call can also be batched, so that we dont needlessly spam the ui-thread with dispatcher calls.

sn4k3:
I will prepare and publish a new PR with it. 
If I manage a way to only have timer running when required I think we should go with DispatcherTimer, because everytime timer ticks is because we need to update in UI Thread, eg: update progress value and/or dismiss. If we have a threaded timer to then always call Dispatcher.Post the benefit is lost.

PR #458, opened 2025-04-06
[TextBox] Fix sizing issues by replacing `StackPanel` with `Grid`

dme-compunet:
Using StackPanel is problematic because even when the parent container provides more space, the TextBox remains at its minimum size due to StackPanel's sizing behavior.

### DIFF
--- a/SukiUI/Theme/TextBoxStyles.xaml
+++ b/SukiUI/Theme/TextBoxStyles.xaml
@@ -134,10 +134,10 @@
 
             <ControlTemplate>
 
-                <StackPanel>
-
+                <Grid RowDefinitions="Auto *">
 
                     <TextBlock Name="floatingWatermark"
+                               Grid.Row="0"
                                Margin="3,5"
                                FontSize="13"
                                FontWeight="DemiBold"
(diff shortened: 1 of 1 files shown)


sn4k3:
`StackPanel` = Bug. I avoid it at all cost...

dme-compunet:
I wouldn't say it's a bug, but it's not suitable for every use case.

sn4k3:
> I wouldn't say it's a bug, but it's not suitable for every use case.

Well, most of the time I use stackpanel I end having problems with clipping and lack of scroll. The panel is so convenient that I guess everyone tries to use it also much easier then set a grid.
I wish stackpanel could be more like a *,* container and not act like having unlimited width/height.

The only safe use for stackpanel is if it have manual size, or a parent with size defined. Else it will always clip content when overflow, even inside a scrollviewer...

dme-compunet:
If the problem is unlimited size, then DockPanel will give you the solution together with `Dock=Top` and `LastChildFill`, so why is StackPanel really needed? A question for thought.. Maybe for its simplicity. I assume that if you create a menu template you will choose StackPanel, not DockPanel.

sn4k3:
I'm not telling that **StackPanel** does not have a place in framework, it certain does, I also use it (with caution) eg: inside Button for a [icon text] content. But its nature have potential to introduce undesired/unknown effect (by the user), mostly because people use it for convenience (Write less code, less verbose, cleaner, easy spacing) when compared to a Grid, that's why I refer to StackPanel as a "bug", because many of app bugs where the layout doesn't adapt are caused by StackPanel.

Recently I submitted a [PR](https://github.com/kikipoulet/SukiUI/pull/371) to fix dialog's not having scrollbar when hosting larger content the viewport, the culprit: StackPanel... So I'm not amazed by see another "bug" here caused by StackPanel, and I expect more to shine...

PR #460, opened 2025-04-09
Add new text and button styles, implement RepeatButton, improve SelectableTextBlock and ToggleButton

sn4k3:
- Added new text styles (Success, Information, Warning, Danger)
  - We need "fixed" styles for basic states such as Success and Error/Danger information
  - I added them to global Light and Dark resources, feel free to adjust the colors as required
  - I plan to extend this to all Buttons in a next PR (Note that current Success and Danger styles are broken/not complete) so we can have (Success, Information, Warning, Danger) buttons, which we cannot rely on color theme, for example, if I opt for a green theme all my flat buttons will be green... Having these fixed styles will facilitate user creating buttons with proper states without much effort.
- `RepeatButton` was lacking style, implemented from `Button`
- Replicate `SelectableTextBlock` from `TextBlock`, improve context menu and set cursor to `IBeam` as per avalonia implementation.

image
image
image

# EDIT 1  - Commit [423352f](https://github.com/kikipoulet/SukiUI/pull/460/commits/423352f3cafc507827bfa2d01ac01d18153f0df9)

- Cleanup and simplify Button template
- Implement the new color styles to buttons (Button, RepeatButton, ToggleButton) -- Success, Information, Warning, Danger
- Fix `ToggleButton` logic, when hover it always show toggle background, leading to not know current state without un-hover first, it now uses the toggle state color and only change BorderBrush on hover.
- Add new `ButtonExtensions.Icon` property to allow inject an Icon into Buttons, we want this for some reasons:
  1. Ease of use and fast inject an icon without ugly and verbose StackPanel + Icon + TextBlock
  2. When using StackPanel + Icon + TextBlock combo, the font size is no longer the default of button, because TextBlock have it own font size and style.
  3. We need icon to match button font size, as any provided icon will constrain up to FontSize, so no more guessing or manual size required.
  - Usage: `<Button suki:ButtonExtensions.Icon="{avalonia:MaterialIconExt Kind=Add}" Classes="Flat" Content="Add" />`

Produces:

Buttons:
image

`ToogleButton`:

https://github.com/user-attachments/assets/c69952a2-74d0-47b9-b0ed-db9144e0918f

# EDIT 2 - Commit [667f3d6](https://github.com/kikipoulet/SukiUI/pull/460/commits/667f3d645c67ef003069465269097716408fdee1)

- Theme `DropDownButton` based on `Button`
- Add `Discrete` button style

Produces:

image

`SplitButton` will be my next victim...


kikipoulet:
I've got a little headache recently so I will take a deeper look later, lots of great ideas here, but just to say that usually I take color inspiration from ant design colors if you want to test with coherent colors with what we usually use:

https://ant.design/docs/spec/colors

Aiming to style all the buttons variants is courageous btw

sn4k3:
> I've got a little headache recently so I will take a deeper look later

I'm also with a flu, and it's raining, so I got some time to tinker with this :D

> just to say that usually I take color inspiration from ant design colors

I think it's best to you to check this out as I'm not the best with colors.
However If you want I can just replicate from this:

image

PR #463, opened 2025-04-13
Create Nightly Release Workflow

sn4k3:
# Nightly Release Workflow

As stated here: https://github.com/kikipoulet/SukiUI/discussions/393
We really need nightly releases as the stable release can take some time.
Some bugs require fast fixes and this can stall an app release/fix (Like my own case)

# How it works

1. Midnight UTC every day:
2. Get semver version from last tag (ATM no tag is available, so I recommend adopting this after publishing the first semver tag)
3. Checks if there are modifications/commits compared with last nightly, if so, trigger the build, otherwise quit.
4. Publish to dotnet using current tag version: `x.x.x-nightlyYYYYmmdd`

# Things to consider

1. Accept this only after first semver tag creation
2. Think if we want this on nuget or choose a separate repository

### DIFF
--- a/.github/workflows/nightly-release.yml
+++ b/.github/workflows/nightly-release.yml
@@ -1,0 +1,92 @@
+name: Nightly Release
+
+on:
+  schedule:
+    - cron: '0 0 * * *'  # Runs at midnight UTC every day
+  workflow_dispatch:  # Allow manual triggering
+
+jobs:
+  check-changes:
+    runs-on: ubuntu-latest
+    outputs:
+      has-changes: ${{ steps.check.outputs.has-changes }}
+      base-version: ${{ steps.get-version.outputs.base-version }}
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0  # Need full history for version checking
+
+      - name: Get last release version
+        id: get-version
+        run: |
+          # Get the most recent release tag matching semantic versioning
+          LAST_TAG=$(git describe --tags --match "[0-9]*.[0-9]*.[0-9]*" --abbrev=0 2>/dev/null || echo "0.0.0")
+          echo "base-version=${LAST_TAG#v}" >> $GITHUB_OUTPUT
+          echo "Using base version: ${LAST_TAG#v}"
+
+      - name: Check for changes since last successful nightly release
+        id: check
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          # Get the date of the last successful nightly release
+          LAST_RELEASE_DATE=$(gh run list --workflow=Nightly\ Release --status=success --limit=1 --json createdAt --jq '.[0].createdAt' 2>/dev/null || true)
+          
+          if [ -z "$LAST_RELEASE_DATE" ]; then
+            echo "No previous successful releases found, proceeding with build"
+            echo "has-changes=true" >> $GITHUB_OUTPUT
+            exit 0
+          fi
+          
+          # Check if there are any commits since last release
+          CHANGES=$(git log --since="$LAST_RELEASE_DATE" --oneline)
+          
+          if [ -z "$CHANGES" ]; then
+            echo "No changes since last release at $LAST_RELEASE_DATE"
+            echo "has-changes=false" >> $GITHUB_OUTPUT
+          else
+            echo "Changes detected since last release at $LAST_RELEASE_DATE"
+            echo "has-changes=true" >> $GITHUB_OUTPUT
+          fi
+
+  publish:
+    needs: check-changes
+    if: needs.check-changes.outputs.has-changes == 'true'
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup .NET
+        uses: actions/setup-dotnet@v4
+        with:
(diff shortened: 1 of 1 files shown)


Insire:
Nightly seems like overkill. Why not just run a CD build with a pre release/alpha version, when something is being pushed to main? And then do non pre release/alpha versions, when a commit has an explicit version tag?

That way everyone can choose what version/commit to use. E.g. if someone were to provide a PR, they could then choose to use the exact pre release package with their fix inside and would not be potentially bothered by other merges.

sn4k3:
> Nightly seems like overkill. Why not just run a CD build with a pre release/alpha version, when something is being pushed to main? 

As explained, my workflow only creates a nightly publish **if** there are changes / new commits relative to the previous nightly. If 10 day has pass but, in that time, nothing was committed there's 0 publishes during that time.

Your suggestion looks more overkill, because *when something is being pushed to main*, in a day if 10 commits are pushes into main, we will get 10 releases?

> And then do non pre release/alpha versions, when a commit has an explicit version tag?

This is already [deployed](https://github.com/kikipoulet/SukiUI/blob/main/.github/workflows/release.yml)

> if someone were to provide a PR, they could then choose to use the exact pre release package with their fix inside and would not be potentially bothered by other merges.

dme-compunet:
I don't think it's a good idea to publish nightly releases to NuGet, it's better to do it in a separate channel like GitHub packages or something else.

sn4k3:
> I don't think it's a good idea to publish nightly releases to NuGet, it's better to do it in a separate channel like GitHub packages or something else.

I agree on that.

> Another thing, it seems that AI wrote a significant part of the workflow (I could be completely wrong), have you had a chance to run it and see that everything works properly?

Was AI generated using the release template of suki. I modified very little. Haven't run to confirm but I believe the workflow will work. I tested some commands like the tag fetcher, and they work. The current models are soo good that I'm amazed to discover resolve on things very new and lacking in information... 
Anyway, I included the line for manual call the workflow under github, it could be tested once approved. If required I can change to use github packages instead.

Insire:
> > Nightly seems like overkill. Why not just run a CD build with a pre release/alpha version, when something is being pushed to main?
> 
> As explained, my workflow only creates a nightly publish **if** there are changes / new commits relative to the previous nightly. If 10 day has pass but, in that time, nothing was committed there's 0 publishes during that time.
> 
> Your suggestion looks more overkill, because _when something is being pushed to main_, in a day if 10 commits are pushes into main, we will get 10 releases? Many commits are not even worth a publish, like formating, space removal, docs, etc...

I guess overkill is not the right word here. My bad. 

Consider the following situation:
- you provide a PR that solves a problem thats blocking you
- there are other PRs already open
- many of them are being merged in a single day (as it seems to happen regularly here)
- your PR is being merged aswell
- but one of them breaks something that you also need

My concern is, that you now can't use that nightly package, that would be unblocking you, despite your PR having been merged.

And yes, that would mean lots of preview nuget packages, as compared to one per day. But is that actually a problem?

Maybe we need 2 kinds of preview nuget packages?
- alpha for stuff from a single PR
- nightly containing everything currently in main?

I'm not fussy about the names, just that we should document our choice of them and what they would mean/contain.

In regards to where to publish them:
- githubs support for nuget was pretty awful last time i checked (which admittedly has been a while since i used it)
- azure devops would be another (free) solution
- i dont see a reason why nuget.org is invalid here. if you dont want to see/use preview packages, then you dont have to

sn4k3:
I understand your point, it's valid but a bit tricky to be perfect.
Imagine the following:

1. A PR is merged and introduces a critical bug for your app (A nuget is created)
2. Your PR which fixes a different problem is also merged (A nuget is created)

You install the nuget from 2., to solve your situation, but you realize you can't use it because of the nasty bug introduced with 1.
Of course, if merge order is inverted it will work... 
The odds on something like this to happen are small.

I'm fine with both ideas but still think nightly are cleaner. In a repo that could submit many individuals commits it will be a spam, at least it should have some criteria.

sn4k3:
Can we work on this? Really need latest patches...

sn4k3:
> This PR will work with the already existing 6.0.2 tag ?

Yes, it will obtain last tag with a version just to build the nightly with the same version, appends "-nightlyDATE", eg: 6.0.2-nightlyYYYYmmdd

sn4k3:
It worked 🚀
However I see two inconveniences

1. "-nightly" is not considerate a "pre-release" under VS nuget. Google says: To mark a NuGet package as pre-release, append a semantic versioning suffix like -alpha, -beta, or -rc to the package version.
2. Current version strategy marks nighty as older version than current stable.

image


To fix:

1. Change "-nightly" to "-preview" (Is what MS uses and seem to work)
2. Use current version and append extra digit, from `6.0.2` to `6.0.2.999` (Like Avalonia does)

Final naming: `6.0.2.999-previewDATE`

What do you think?

sn4k3:
Avalonia does that for nuget repo, however for nightly they just set PATH to static 999

image

> That's not true, it is indeed a candidate as "prerelease"

Strange, "yesterday" the package shows normally with "pre-releases" unchecked but today is works only with "pre-releases" checked. No clue what happened with VS. As it's working now I kinda like the "nightly" suffix the most.

So far the options are:

- x.x.x.999-previewYYYYmmdd
- x.x.(x+1)-previewYYYYmmdd

Both represent a version ahead current.
I guess the vote goes for the 2nd convention?

PR #464, opened 2025-04-14
Modify the Checkbox indeterminate style to be similar to checked

YJammak:
Modify the Checkbox indeterminate style to be similar to checked
Snipaste_2025-04-14_11-32-38


sn4k3:
Note that a converter is not required, there is a built-in expression: `!!Variable`

`{Binding !!IsChecked, RelativeSource={RelativeSource TemplatedParent}}`

YJammak:
> Note that a converter is not required, there is a built-in expression: `!!Variable` this is exactly the `Not False` / `!false`
> 
> `{Binding !!IsChecked, RelativeSource={RelativeSource TemplatedParent}}`

I tried it, but it doesn't work when IsChecked is null.

sn4k3:
> I tried it, but it doesn't work when IsChecked is null.

Ah, yes, you right, this needs a converter because you want `Null` as `True`... When I first saw I thought you want `Null` as `False`, my bad...

PR #465, opened 2025-04-14
SplitButton style and many others improvements

sn4k3:
This is the continuation of my previous [PR](https://github.com/kikipoulet/SukiUI/pull/460)

- Complete the `SplitButton` style
- Simplify and make sure all buttons use same styling without redudancy
- Every button that has a flyout open will get represented with it pointover styles
- Add `Small` class to buttons, useful to put just a icon for an action
- Improve `DropDownButton` Margin related to the glyph
- Implement colors to `ProgressBar`
- Implement colors to `ToggleSwitch`
- Remove some control styles and convert to ResourceDictionary

Produces:

image

image

image

image

https://github.com/user-attachments/assets/baa16282-dfce-4275-9dc0-1c880b4465f7


sn4k3:
Thanks for accept my PRs, the library is now richer.

> This little vertical bar next to the arrow is really elegant, great !

I always hated the default broken avalonia design for SplitButton, but they have a separator. 
As a split button there should be an indicator of a division (two distinct components) that is clear for the user.
I come with that vertical bar and thought it would look awesome and modern :)

sn4k3:
I also saw your color changes.
My observations:

- Success have no tone differentiation on light theme:

```xml
    <Color x:Key="SukiSuccessColor">#279f27</Color>
    <Color x:Key="SukiSuccessMediumColor">#279f27</Color>
    <Color x:Key="SukiSuccessDarkColor">#279f27</Color>
```

- On both themes, the warning is too deep orange... Easy mistaken with red, we should use easy colors for people with trouble in differentiating tones, see bellow and focus the last button, does it look warning? It's more danger alike:

image

sn4k3:
> it has to not clash with the overall ambiance of SukiUI

Yes I agree, that's why we need to meet that two factors together (Ambiance of SukiUI + The iconic color that auto tell us Warning)

That new warning color looks much better, for me its suitable candidate.

> playing with opacity

By opacity you really mean the button opacity?
Or background alpha component?

If opacity, is a bad idea as it will opaque everything inside button.
If background alpha (Argb) then it's a good ideaa, but also the only way to have a single color among light and dark theme that really works due the background, plus it will shade the color based on current background which can be cool...
Alpha is in fact what ant design show us:

image

PR #468, opened 2025-04-19
[Feature] Add async blocking Dialogs

Insire:
**What**
MessageBoxes in Winforms and WPF generally block execution, until the user deals with them in some manner. SukiUI doesnt support that yet. I added 2 overloads to ``FluentSukiDialogBuilder`` that adds a way to asynchronously wait for them to be closed, before the calling code can continue. You also get a way to figure out, which button was clicked, if one was clicked.

**What else?**
I restructured the DialogsView and the corresponding Viewmodel, so that its easier to find things.

**Bugs?**
Nothing new, but i noticed dialogs are flicking when they are being opened via ``SukiMessageBox.ShowDialog(new SukiMessageBoxHost());`` in the Demo app. So that probably needs fixing.


sn4k3:
> but i noticed dialogs are flicking when they are being opened via SukiMessageBox.ShowDialog(new SukiMessageBoxHost());

I notice that when I made the implementation. I checked but can't find a valid cause. The strange thing is that happen in some dialog but not others. For example, it does with the most basic ones like "Question" but with "Long" or "Animated" doesn't. Also others that blink not always blink, it's indeed strange.

Insire:
> > but i noticed dialogs are flicking when they are being opened via SukiMessageBox.ShowDialog(new SukiMessageBoxHost());
> 
> I notice that when I made the implementation. I checked but can't find a valid cause. The strange thing is that happen in some dialog but not others. For example, it does with the most basic ones like "Question" but with "Long" or "Animated" doesn't. Also others that blink not always blink, it's indeed strange because nothing is changed after the Windows is shown, so I don't have an explanation for such artifact.

If i had to make an educated guess, then i'd say its because of the new instance of ``SukiMessageBoxHost`` thats being created there and it being a (new) control. Ideally you'd keep the existing instance and update its configuration, or what the reason is for providing a new instance.

sn4k3:
`because of the new instance of SukiMessageBoxHost`

That doesn't make sense because control is given to window before it gets draw, this yield same effect if I write a xaml Window with a control inside it. I tested using a normal `Window` instead of `SukiWindow` and does not reproduce the problem. My guess will go to animations. I will try to use the preview of avalonia and see if that helps (New preview much smoother)

kon3ko:
> `because of the new instance of SukiMessageBoxHost`
> 
> That doesn't make sense because control is given to window before it gets draw/Loaded, this yield same effect if I write a xaml Window with a control inside it. I tested using a normal `Window` instead of `SukiWindow` and does not reproduce the problem. My guess will go to animations. I will try to use the preview of avalonia and see if that helps (New preview much smoother)
> 
> **EDIT:** New avalonia preview does the same flicker.

How do you solve this problem now?

PR #469, opened 2025-04-19
[Performance] Remove unnecessary allocations in `FluentAnimator<T>`

dme-compunet:
- Convert `FluentAnimator<T>` to `ref struct` since it only lives on the stack
- Use global `CubicEaseInOut` instead of creating new instances per animation
- Replace `CancellationTokenSource` return with `WithCancellationToken` method to avoid unnecessary allocations when cancellation isn't needed
- Add ability to run animation without setting From value - uses current value as starting point
- Add `RunAsync` method to allow awaiting animation completion

### DIFF
--- a/SukiUI/Helpers/FluentAnimator.cs
+++ b/SukiUI/Helpers/FluentAnimator.cs
@@ -1,97 +1,89 @@
-﻿using Avalonia.Styling;
-using Avalonia;
+﻿using Avalonia;
 using Avalonia.Animation;
 using Avalonia.Animation.Easings;
-using System;
-using System.Threading;
+using Avalonia.Styling;
 
 namespace SukiUI.Helpers;
 
+/*
+ * _myButton.Animate(OpacityProperty)
+ *          .From(0).To(1)
+ *          .WithDuration(TimeSpan.FromMilliseconds(1000))
+ *          .WithEasing(new CubicEaseOut())
+ *          .Start();
+ */
 
 public static class FluentAnimatorExtensions
 {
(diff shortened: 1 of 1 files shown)


Insire:
This is looking nice. 

I noticed that there seems to be a binary breaking change here. The return value of `Start` has changed from `CancellationTokenSource` to ``void`. Which makes sense, but in the off chance someone was using it, it needs to be documented somewhere. So that there are no suprises when upgrading.

@kikipoulet do we have a changelog somewhere or maybe a migration guide to gather these kinds of things?

kikipoulet:
> This is looking nice. 
> 
> I noticed that there seems to be a binary breaking change here. The return value of `Start` has changed from `CancellationTokenSource` to ``void`. Which makes sense, but in the off chance someone was using it, it needs to be documented somewhere. So that there are no suprises when upgrading.
> 
> @kikipoulet do we have a changelog somewhere or maybe a migration guide to gather these kinds of things?

This PR is a super initiative !

I'm not worried about that change specifically because it's a very recent api I made that is not even in the documentation at all, it's even a kind of WIP in a way. It's been few weeks I try to : 
1. Find ways to create animations more easily, at least imo (thanks @dme-compunet for polishing!)
2. experiment to create a sort of "organic" animations to make the library feels more alive in the future.

But I see that recently the library get a kind of second wave of interest so a changelog would be a good idea anyway because I feel we will need it.

dme-compunet:
The text here needs to be changed, but I prefer that @kikipoulet select the change.

https://github.com/kikipoulet/SukiUI/blob/3eea57ace9169b6597bb5117bb944d226eb6ec11/SukiUI.Demo/Features/Helpers/HelpersView.axaml#L280-L282

PR #479, opened 2025-05-08
About Textbox Height

futurex894:
Setting 'RowDefinitions="Auto Auto"' will make it impossible to set the height of the TextBox. Moreover, the TextBox won't be able to expand within the Grid, and the scrollbar of the TextBox will also stop functioning.

### DIFF
--- a/SukiUI/Theme/TextBoxStyles.xaml
+++ b/SukiUI/Theme/TextBoxStyles.xaml
@@ -134,7 +134,7 @@
 
             <ControlTemplate>
 
-                <Grid RowDefinitions="Auto Auto">
+                <Grid RowDefinitions="Auto *">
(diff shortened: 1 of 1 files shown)


futurex894:
> Then it takes too much height by default.
> 
> {6B78D1D7-64FA-4857-8071-308C2167E450}
Yes, you are right.
image
But I think we can modify the MinHeight to achieve the same effect.
image
image
If you set it as "RowDefinitions = "Auto Auto"", it will lead to the following error.
image
image
image

sn4k3:
Can also confirm that current implementation of TextBox have bug of cut vertical content (No scroll bar created)
All my TextBox are now absent of a vertical scrollbar.

image

PR #480, opened 2025-05-10
SukiBackground optimize

DearVa:
cdb9d65: SukiBackground now can share scale and translate (Behaves like ExperimentalAcrylicBorder). You can ignore this if you don't want it.
de46b58: Add CornerRadius for SukiBackground
e484bed: fix nullable


sn4k3:
Can review this and update with main?
Also can describe a little futher what was been done and why / benefits?

PR #485, opened 2025-05-14
Fix Windows cut in Maximize mode 

### LINKED ISSUE #484: [Bug] Windows cut in Maximize mode
### Check the following items

### Description of the issue

Normal mode: (Look ok)
Image

Maximize mode: (Margins cropped)

Image

Fullscreen mode: (Looks ok)
Image

There is also a bug in fullscreen mode, the auto hide title bar does not display, because minimum Y coord that app can obtain is 7 which never trigger the `if (position.Y <= 3)` it was working before when implemented, not sure where the regression is. But can be fixed by using `10` at if condition

### Package Version

6.0.2

### Environment

Windows 11

### Expected Behavior

Windows to not cut content when maximized (Looks like to be 7px)

### Reproduction

- Maximize demo window. Notice the cut borders.

### Additional Information

_No response_

sn4k3:
Fixes #484

- Allow set Margin and Padding for `SukiWindow`

### DIFF
--- a/SukiUI/Controls/SukiWindow.axaml
+++ b/SukiUI/Controls/SukiWindow.axaml
@@ -26,7 +26,7 @@
         <Setter Property="FlowDirection" Value="{DynamicResource FlowDirectionPrimary}" />
         <Setter Property="Template">
             <ControlTemplate>
-                <Panel Name="PART_Root">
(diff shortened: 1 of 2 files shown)


CTimet:
I meet a same issues. Hope this can be fixed

PR #494, opened 2025-06-08
Rework TextBox and NumericUpDown

sn4k3:
- Convert `TextBox` and `NumericUpDown` to a `ResourceDictionary`
- Bring back the floating watermark (It was removed somehow), using avalonia implementation but never hide floating if text is empty.
- Fix reveal and delete utilities classes
- Fix some missing alignments 
- Fix some paddings
- Fix `NumericUpDown` content getting cut
- Fix `NumericUpDown` unable to set foreground
- Fix `NumericUpDown` unable to use Left button spinner
- Change `Prefix` and `Unit` from `string` to `object?`, this will allow to use controls, such as icons

image
image
image
image


kikipoulet:
Please tell me it doesn't change the original size/scaling of the text box 👋

sn4k3:
> Please tell me it doesn't change the original size/scaling of the text box 👋

I adjusted padding to make it uniform, that lead to a 2px difference in height, from 36px to 38px

image

I can make it 36px if you prefer it.

PR #510, opened 2025-08-01
Added SukiSideMenu customization features

### LINKED ISSUE #496: [Feat] SukiSideMenu Support Manually Content Set

- [x] I have searched for duplicate issues.

The current situation is that when the program loads, all Viewmodels will be initialized. My plan is to decorate the Viewmodel base class, generate a specific binding class used to NavigatorMenu, and then navigate to the real viewModel implementation class through the binding class to achieve delayed loading for improve performance. My current approach is to manually set the PART_TransitioneContentControl content of SideMenu through reflection, which is obviously not flexible enough and sometimes conflicts with the internal switching logic of SideMenu. Can we add a manually set switch so that we can implement our own content switching logic?

### Package Version

6.0.2

deepea:
Added some things to make customizing the SukiSideMenu easier:

- Manually set content on SukiSideMenu

For those that prefer using a more standard data binding approach to displaying the content. By setting the `UseCustomContent` property to True and providing view data to the `Content` property, SukiSideMenu will override the default behavior and instead use said content instead of looking to the `PageContent` property of the SukiSideMenuItem.

Resolves #496 

- Provide an additional display area on menu items

Since the menu item header bind texts directly to a TextBlock I felt that adding in a new spot for customization would be the simplest backwards-compatible way to offer some addition functionality. Just provide a data template in the `SideDisplayTemplate` property. See test examples with an icon, a context menu, text.

menu-sides
menu-sides-ctx
menu-sides-submenus
menu-collapsed

- Made logic methods protected virtual

Allows easy overriding of logic in derived classes. e.g. Adding search operators to the filter, changing how displaying content works, etc.

### DIFF
--- a/SukiUI/Controls/SukiSideMenu.axaml.cs
+++ b/SukiUI/Controls/SukiSideMenu.axaml.cs
@@ -132,6 +132,24 @@ public class SukiSideMenu : TreeView
         set => SetValue(FooterContentProperty, value);
     }
 
+    public static readonly StyledProperty<object?> ContentProperty =
+        AvaloniaProperty.Register<SukiSideMenu, object?>(nameof(Content));
+
+    public object? Content
+    {
+        get => GetValue(ContentProperty);
+        set => SetValue(ContentProperty, value);
+    }
+
+    public static readonly StyledProperty<bool> UseCustomContentProperty =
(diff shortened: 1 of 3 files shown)


cditqq:
Is there a usage example available?

Sfayx:
great work!

PR #514, opened 2025-08-07
feat: Unify foreground color for RadioButton.Chips on checked state

KaranocaVe:
This pull request introduces a new style for checked radio buttons with the `Chips` style, improving UI consistency by ensuring the foreground color is set to white when selected.

RadioButton style updates:

* Added a style for `RadioButton.Chips:checked` to set the `Foreground` property to white, ensuring selected chips are visually distinct.

image

### DIFF
--- a/SukiUI/Theme/ListBoxStyles.xaml
+++ b/SukiUI/Theme/ListBoxStyles.xaml
@@ -26,7 +26,6 @@
         </Border>
     </Design.PreviewWith>
     <Style Selector="ListBox">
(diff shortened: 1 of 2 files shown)


KaranocaVe:
**ListBox Style Updates:**
* Removed the default foreground color (`TextBlock.Foreground`) setter from the `ListBox` style in `ListBoxStyles.xaml` to allow more flexibility or to inherit the foreground color from elsewhere.

First, removing this won't affect the foreground color of the internal TextBlock because global rules exist. Second, the settings here would override the foreground color of Text*.Foreground internally.

PR #515, opened 2025-08-08
fix: Adjust margin for ComboBox styles to improve layout

KaranocaVe:
This is a bug.
Before:
image

After:
image


sirdoombox:
I actually just tripped over this myself in something I'm working on. Don't know how it went unnoticed for this long.

PR #606, opened 2026-04-22
Project Setup Tweaks

artehe:
So all the PRs I'd done up to now should be fully compatible with Avalonia 11, this does a few little things which are somewhat breaking to the current setup. 

Feel free to reject this or have it split into little pieces. However after this it should be somewhat simple in what is required to support Avalonia 12.

- Use just one physical version of the logo and then link to it (avoids having
- Use latest Avalonia 11 version
- Change DotNet version to 8 (minimum required for Avalonia 12)
- Remove System.Text.Json as it is contained within the new dotnet version
-  Enable Trim analyzer and trimming (Note: this isn't fully supported yet but should highlight areas that need work to have it working fully)


sn4k3:
If upgrading to NET 8.0, System.Linq.Dynamic.Core dep should also be removed.

artehe:
> If upgrading to NET 8.0, System.Linq.Dynamic.Core dep should also be removed.

You're right, though I was gonna leave that initially cause it required reworking/removing these two files and I was reluctant to touch them just yet:
- [SukiUI/Helpers/ConditionalXAML/InlineSharp.cs](https://github.com/kikipoulet/SukiUI/blob/main/SukiUI/Helpers/ConditionalXAML/If.cs)
- [SukiUI/Helpers/ConditionalXAML/InlineSharp.cs](https://github.com/kikipoulet/SukiUI/blob/main/SukiUI/Helpers/ConditionalXAML/InlineSharp.cs)

sn4k3:
I've reviewed and tested. Looks good 👍

PR #608, opened 2026-04-29
Prevent context menu following cursor on open

### LINKED ISSUE #585: [Bug] The right-click menu will follow the cursor when it pops up.
### Check the following items

### Description of the issue

Image

### Package Version

6.0.3

### Environment

Windows 11 26200

### Expected Behavior

Do not let it move with the cursor when it pops up.

### Reproduction

The right-click menu will follow the cursor when it pops up.

### Additional Information

_No response_

artehe:
Should resolve #585 so the context menu won't follow the cursor while the open animation is running.

### DIFF
--- a/SukiUI/Theme/ContextMenu.axaml
+++ b/SukiUI/Theme/ContextMenu.axaml
@@ -7,9 +7,13 @@
         <Setter Property="Template">
             <ControlTemplate>
                 <!--  Tiny margin helps prevent mouse-over immediately  -->
-                <LayoutTransformControl Name="PART_LayoutTransform"
-                                        Margin="1,0,0,0"
-                                        RenderTransformOrigin="0%, 0%">
+                <Border Name="PART_LayoutTransform"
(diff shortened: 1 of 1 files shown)


sn4k3:
It's better but still have the problem if you fast to move your mouse:

image

artehe:
> It's better but still have the problem if you fast to move your mouse:
> image

Ye I see what you mean, although I am unsure what to do about that.

Best I can think of at the moment is for each control that can have a context menu adding code to capture the pointer when the context menu is opened so we can fix it to that location. Although I don't think that this is very practical 😢 

I'm open to ideas.

sn4k3:
Maybe we can change the effect, idk... For now I'm accepting this as it already better than before.

artehe:
> Maybe we can change the effect, idk... For now I'm accepting this as it already better than before.

Ye the animation exposes the issue more, in theory I think this would happen with default context menu opening (if you where quick enough with the mouse).

PR #615, opened 2026-06-01
Fix: The context menu does not disappear when clicking on an empty area

deepea:
Fixes #613

PR #618, opened 2026-06-23
Fix titlebar context menu dismissal

atticus-lv:
Preserves the context-menu dismissal fix from #615 and removes the old SukiWindow titlebar hack


sn4k3:
Problem come from av12 upgrade, regression in https://github.com/atticus-lv/SukiUI/commit/a8e07126a0cf18b51074a11467157e14040676c8

PR #634, opened 2026-08-19
perf: eliminate repeated renderer and editor setup

KaranocaVe:
## Scope

This PR removes repeated runtime-effect and Demo editor setup work while preserving rendered output and animation timing.

- Cache `SukiEffect.FromString` instances by exact source under a lifecycle lock.
- Precompute the existing source-based `SukiEffect` hash.
- Cache `SKShader` instances only for non-animated standard-uniform effects, keyed by effect, bounds, and alpha.
- Reuse draw-state objects and dispose cached native resources when the visual, effect, theme, or bounds change.
- Keep animated and custom-uniform effects on their original per-render uniform path so time-dependent output is not quantized or reused incorrectly.
- Reuse immutable TextMate grammar registries and avoid rebuilding unused registries when a `CodeEditor` is detached and reattached during a page transition.
- Cache the immutable dynamic-blur clamp effect and dispose temporary native runtime-effect uniform data on every draw.
- Replace the Demo Dock explorer's synchronous parent-directory scan with a bounded deterministic demo tree.
- Add automated Desktop Demo interaction coverage for dynamic blur scrolling, Squishy dragging, page transitions, the bounded Dock tree, embedded shader compilation, and CodeEditor reattachment.

## Rendering fidelity

- `cells.sksl` is byte-identical to `origin/main` (SHA-256 `3f9158a4720be4a442272d7fedfa791346da69161e8ef325e89deb7c0173b8d2`).
- `waves.sksl` keeps its original `x - y * floor(x / y)` calculation and call site. Only the user-defined function name changes from `mod` to `sukiMod` to avoid a collision with the SkSL intrinsic.
- Animated shader invalidation is display-rate again; no shader-time bucketing remains.
- Transition alpha remains a shader uniform, and the Side Menu keeps its original outgoing-content fade.
- Dynamic blur keeps its original refresh behavior, blur parameters, and output; only its immutable compiled clamp effect and per-draw temporary native objects now have explicit reuse/disposal lifecycles.
- CodeEditor setup stays at the original construction point; the change only removes later registry allocations whose result was not used by the installed TextMate instance.

## CodeEditor reattachment measurement

The new automated Headless Desktop test alternates an XML `CodeEditor` with another control for 12 warm-up frames and records 40 reattachments. On the same local machine, comparing `d9d6f8c` with `d8805be` (included in the current head):

| Metric | Before | After |
| --- | ---: | ---: |
| Frame time p50 / p95 / max | 4.763 / 7.081 / 8.288 ms | 0.388 / 1.318 / 1.465 ms |
| Allocation p50 / p95 / max | 3.881 / 3.964 / 3.969 MB | 0.027 / 0.113 / 0.113 MB |

A real Desktop Demo `Effects -> Expander -> Effects -> Expander` probe confirms the allocation reduction on the UI thread: the first Effects selection changed from 9.22 MB to 5.37 MB, and the cached return changed from 4.17 MB to 0.31 MB. Wall-clock navigation samples remain sensitive to warm-up, GC, and desktop scheduling, so this PR does not claim a whole-Demo navigation p95 improvement from this change alone.

## Dynamic blur native lifecycle

`SKRuntimeEffectUniforms` owns an `SKData` allocation. Dynamic blur created one on every draw without deterministic disposal, and could recompile its immutable clamp effect when its draw operation was recreated. The renderer now scopes temporary objects to one draw and shares the compiled clamp effect until process exit. The real 300-frame Metal dynamic-blur scroll probe remains smooth after the change: p50/p95/max 16.249/16.986/20.916 ms, 0 missed 60 Hz frames, `gpu-blur=True`. This confirms the GPU path and output path remain healthy; no before/after frame-rate percentage is claimed for this lifecycle correction.

## Earlier steady-state Desktop Demo context

An earlier clean 300-frame comparison between `d628695` and `d72aba5` used macOS 26.6.2, Apple M5 (10-core GPU), Metal, a 1920x1080 display at 2.0 scaling, and the real Desktop Demo window (1064x647 logical). Each scenario had a 90-frame warm-up. `gpu-blur=True` confirmed the GPU path.

| Scenario | Before p50 / p95 / max | After p50 / p95 / max | Missed 60 Hz frames |
| --- | ---: | ---: | ---: |
| Progress page scroll | 16.171 / 16.750 / 17.148 ms | 16.200 / 16.909 / 17.173 ms | 0 / 0 |
| Squishy drag | 16.181 / 16.825 / 17.217 ms | 16.188 / 16.924 / 17.224 ms | 0 / 0 |

The 0.10-0.16 ms p95 differences are within short-run desktop scheduling noise; this PR does not claim an end-to-end frame-rate improvement for these already smooth steady-state interactions. Earlier Cells, animation-bucket, and paint-alpha benchmark claims were removed because those output-altering implementations are no longer present.

Cold navigation remains a separate issue: large synchronous XAML/control-tree creation and layout are outside the static shader-cache path. The PR now has a regression measurement for repeated CodeEditor attachment, while broader page-tree work remains under investigation.

## Validation

- `dotnet build Suki.sln --no-restore -c Release -clp:ErrorsOnly` -- passed, 0 errors.
- `dotnet test SukiUI.PerformanceTests/SukiUI.PerformanceTests.csproj -c Release --no-restore --logger 'console;verbosity=detailed' -clp:ErrorsOnly` -- passed, 7/7.
- `git diff --check` -- passed.
- Both `cells` and `waves` compile as embedded background shaders.


SukiUI/Utilities/Effects/EffectDrawBase.cs
```diff
@@ -15,13 +16,18 @@ namespace SukiUI.Utilities.Effects
     public abstract class EffectDrawBase : CompositionCustomVisualHandler
     {
         public static readonly object StartAnimations = new(), StopAnimations = new(), 
-            EnableForceSoftwareRendering = new(), DisableForceSoftwareRendering = new();
+            EnableForceSoftwareRendering = new(), DisableForceSoftwareRendering = new(),
+            DisposeHandler = new();
+
+        private const int AnimatedShaderFramesPerSecond = 30;
```
sn4k3: The FPS are fixed and no way to configre, there are no benefit in going lower/higher FPS?

sn4k3:
Thank you for the contribution :)
I've made some changes, please re-test and compare with your benchmark:

1. RemoveStaleAnimatedShaders became PruneShaderCache(timeBucket, width, height), now called on every cache miss rather than only when the time bucket rolls over. It evicts on either condition: a different time bucket for animated keys, or a different render size for any key. That kills the resize accumulation for the non-animated default path while leaving the animated eviction behavior intact.

2. if (_isDisposed) return; at the top of OnRender, which is the single funnel into both Render and RenderSoftware, so all three derived handlers are covered without touching them.

3. the private const became a protected property with the constant as its default, clamped to ≥1 and flushing the cache on change. Raising it to 120 gets near-real-time motion at the cost of more compiles. I did not add a way to disable caching entirely: callers no longer dispose the shaders they receive, so an uncached path would leak. The property is the opt-out that fits the current ownership model.

4. AddCachedShader now takes and returns SKShader? and returns early on null, so a failed ToShader neither poisons the cache nor throws during invalidation.

5. the outgoing shader is built inside the lerped < 1 branch, so the transition's final frame no longer compiles one just to discard it. This also removes the window where a disposed managed SKShader sat assigned to _oldEffectPaint.

6. InvalidateShaderCache(SukiEffect) returns early when the target equals the live _effect. Those entries are genuinely still valid for the current effect, so keeping them is correct rather than just a workaround.

KaranocaVe:
> Thank you for the contribution :) I've made some changes, please re-test and compare with your benchmark:
> 
> 1. RemoveStaleAnimatedShaders became PruneShaderCache(timeBucket, width, height), now called on every cache miss rather than only when the time bucket rolls over. It evicts on either condition: a different time bucket for animated keys, or a different render size for any key. That kills the resize accumulation for the non-animated default path while leaving the animated eviction behavior intact.
> 2. if (_isDisposed) return; at the top of OnRender, which is the single funnel into both Render and RenderSoftware, so all three derived handlers are covered without touching them.
> 3. the private const became a protected property with the constant as its default, clamped to ≥1 and flushing the cache on change. Raising it to 120 gets near-real-time motion at the cost of more compiles. I did not add a way to disable caching entirely: callers no longer dispose the shaders they receive, so an uncached path would leak. The property is the opt-out that fits the current ownership model.
> 4. AddCachedShader now takes and returns SKShader? and returns early on null, so a failed ToShader neither poisons the cache nor throws during invalidation.
> 5. the outgoing shader is built inside the lerped < 1 branch, so the transition's final frame no longer compiles one just to discard it. This also removes the window where a disposed managed SKShader sat assigned to _oldEffectPaint.
> 6. InvalidateShaderCache(SukiEffect) returns early when the target equals the live _effect. Those entries are genuinely still valid for the current effect, so keeping them is correct rather than just a workaround.

Thanks for the lifecycle hardening. I synchronized `191102e`, re-ran the checks and probes, and pushed one small follow-up commit (`4a5483a`) that precomputes the existing source-based `SukiEffect` hash. It preserves the current equality semantics while avoiding a full SKSL-string hash on every shader-cache lookup.

- 10,000 direct shader creations: 28.23–29.74 ms and about 16.04 MB allocated.
- 10,000 hits through the production cache path: 2.29–2.40 ms and 128 B allocated.
- The cache retained one entry after 500 distinct resize sizes and one entry after 40 animated time buckets.
- The precomputed hash reduced 10,000 `SukiEffect.GetHashCode()` calls from 22.14–22.52 ms to 2.39–2.57 ms.

I also re-ran the Metal probe on the Apple M5. The synchronized `Flush + Submit(true)` wall time still varied with warm-up and GPU clock state, so I removed the earlier fixed GPU-percentage claim from the PR description. The updated description retains the reproducible CPU allocation, cache-bound, and local Metal observations without presenting them as a stable FPS result.

KaranocaVe:
Still investigating performance issue

KaranocaVe:
Follow-up pushed in aa19bf1 and b8d9d6c.

The remaining performance paths are covered: bucket-aligned animation invalidation, transition alpha moved out of the shader cache, and deterministic disposal of runtime-effect uniforms and children. SukiUI/Content/Shaders/Background/cells.sksl now measures 67.0-67.7% lower Metal wall time in the controlled probe, and SukiUI/Content/Shaders/Background/waves.sksl compiles.

The PR description now includes the full before/after protocol, CPU allocation results, Metal observations, validation, and the transition-compositing caveat.

SukiUI/Controls/GlassMorphism/BlurBackground.cs
```diff
@@ -20,24 +20,29 @@ public bool IsDynamic
...
     {
         get => GetValue(IntensityFactorProperty);
         set => SetValue(IntensityFactorProperty, value);
     }
-    
+
     static BlurBackground()
     {
         AffectsRender<BlurBackground>(IsDynamicProperty, IntensityFactorProperty);
+        AppDomain.CurrentDomain.ProcessExit += (_, _) =>
```
sn4k3: Is this really necessary? Will not a app close release all memory?

KaranocaVe:
Follow-up: `827ceb1` removes repeated static work from the published `RadialGauge` control. It reuses the needle transform and preserves segment geometry across `Value` updates; geometry is invalidated for range, angle, rim-thickness, collection, segment, and bounds changes. No shader math, frame cadence, or visual parameter is changed.

A new Release Headless control-level probe uses a 360x360 gauge with 24 segments and 24 ticks, 12 warm-up frames, then 40 `Value` updates at 2.0 render scaling. The comparison is `74b681a` versus this commit on the same machine:

| Metric | Before | After | Change |
| --- | ---: | ---: | ---: |
| Frame time p50 / p95 | 0.737 / 0.887 ms | 0.130 / 0.151 ms | -82.4% / -83.0% |
| Allocation p50 / p95 | 548,648 / 548,664 B | 37,720 / 37,736 B | -93.1% / -93.1% |

The regression test also verifies that segment geometry is reused for `Value` updates and rebuilt after range or bounds changes, while the displayed final value remains correct. `dotnet build Suki.sln -c Release --no-restore -clp:ErrorsOnly` and `dotnet test SukiUI.PerformanceTests/SukiUI.PerformanceTests.csproj -c Release --no-build` both pass (8/8).

I also attempted a fresh real Desktop Metal run after the change. Avalonia.Native fails before compositor initialization because the current desktop session cannot start its RenderTimer (`-6661`), so this follow-up intentionally makes no new GPU performance claim.

KaranocaVe:
GPU retest correction: the Desktop RenderTimer is available again. I reran the control change in a real macOS Desktop session at 2.0 scaling, not Headless. The identical temporary pressure harness contained 24 `RadialGauge` controls, each with 24 segments and 24 ticks; it updated all values for 90 warm-up frames and 300 measured frames. The harness was removed after measurement and is not part of this PR.

| Run | `74b681a` p50 / p95 / max | `827ceb1` p50 / p95 / max | Missed 60 Hz frames |
| --- | ---: | ---: | ---: |
| 1 | 16.351 / 25.828 / 59.010 ms | 16.065 / 16.759 / 17.104 ms | 18 -> 0 |
| 2 | 16.180 / 25.537 / 43.538 ms | 16.070 / 16.771 / 19.659 ms | 16 -> 0 |

Across both runs, p95 fell from 25.54-25.83 ms to 16.76-16.77 ms (about 34-35%), and the repeated long frames disappeared. This is the real-rendering complement to the allocation benchmark; it is specific to repeated `RadialGauge.Value` updates under a deliberately dense control workload.

KaranocaVe:
The RadialGauge optimization and its measurements have been split into #639. This PR now ends at 74b681a and no longer contains that control change; please review the gauge work independently.

KaranocaVe:
need more time to refine the optimization
and may divide it to more small pr

PR #636, opened 2026-08-20
Remove reflection based calls from enum extentions.

### LINKED ISSUE #529: [Bug] AOT Incompatible Reflection
### Check the following items

### Description of the issue

For toast Suki is performing reflection based enumeration for
'SukiUI.Enums.SukiButtonStyles[]':"

```
  //
    // Summary:
    //     Gets all the set flags of an enum value(s).
    //
    // Parameters:
    //   flags:
    //
    //   ignoreFlag:
    //     Set a flag to exclude from returning, eg: If you want to ignore or remove an
    //     always return 0 flag.
    //
    // Type parameters:
    //   T:
    //
    // Remarks:
    //     For enums with System.FlagsAttribute.
    //     If you have an enum value set to 0 it will always return. (Filter it before or
    //     after calling this function)
    //     If you have negative value it can return all flags as undesired effect.
    public static IEnumerable<T> GetSetFlagsIgnoring<T>(this T flags, T ignoreFlag) where T : Enum
    {
        long flagsInt64 = Convert.ToInt64(flags);
        long num = Convert.ToInt64(ignoreFlag);
        flagsInt64 &= ~num;
        foreach (T value in Enum.GetValues(typeof(T)))
        {
            long num2 = Convert.ToInt64(value);
            if ((flagsInt64 & num2) == num2)
            {
                yield return value;
            }
        }
    }
```

```
 IEnumerable<string> items = from v in style.GetSetFlagsIgnoring(SukiButtonStyles.Standard)
                                        select v.ToString();
```

Least to say this will break AOT compiled / WASM builds

### Package Version

6.0.3-nightly20250810

### Environment

Windows 11

### Expected Behavior

Doesn't rely on reflection

### Reproduction

Build an app AOT

### Additional Information

_No response_

artehe:
Resolve #529 by switching to use ```Enum.GetValues<T>()``` rather than ```Enum.GetValues(typeof(T))``` within EnumExtensions.cs

### DIFF
--- a/SukiUI/Extensions/EnumExtensions.cs
+++ b/SukiUI/Extensions/EnumExtensions.cs
@@ -8,9 +8,9 @@ public static class EnumExtensions
     /// <typeparam name="T"></typeparam>
     /// <param name="orderByName">True to order by enum name, otherwise it will order was defined on enum.</param>
     /// <returns></returns>
-    public static IEnumerable<T> GetValues<T>(bool orderByName = false) where T : Enum
+    public static IEnumerable<T> GetValues<T>(bool orderByName = false) where T : struct, Enum
     {
-        var values = (T[])Enum.GetValues(typeof(T));
+        T[] values = Enum.GetValues<T>();
         return orderByName
             ? values.OrderBy(e => e.ToString(), StringComparer.Ordinal)
             : values;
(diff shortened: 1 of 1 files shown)


sn4k3:
Thank you :)

PR #639, opened 2026-08-20
perf: avoid rebuilding radial gauge geometry

KaranocaVe:
## Summary

`RadialGauge.Value` changes no longer rebuild segment paths, brushes, or the needle transform when their inputs are unchanged.

- Reuse the needle `RotateTransform`.
- Reuse static segment geometry across value-only updates.
- Invalidate that geometry when range, angle, rim thickness, segment data, collection, or bounds change.
- Preserve the existing value calculation, visual parameters, and update cadence.

## Validation

- `dotnet build Suki.sln -c Release -clp:ErrorsOnly` passed.
- `git diff --check` passed.

## Performance

The control change was measured before this PR was split, using the same `RadialGauge` implementation and a real macOS Desktop rendering session at 2.0 scaling. The temporary pressure harness rendered 24 gauges, each with 24 segments and 24 ticks, updated values for 90 warm-up frames, then measured 300 frames. The harness is not included in this PR.

| Run | Previous implementation p50 / p95 / max | This implementation p50 / p95 / max | Missed 60 Hz frames |
| --- | ---: | ---: | ---: |
| 1 | 16.351 / 25.828 / 59.010 ms | 16.065 / 16.759 / 17.104 ms | 18 -> 0 |
| 2 | 16.180 / 25.537 / 43.538 ms | 16.070 / 16.771 / 19.659 ms | 16 -> 0 |

The p95 interval dropped by about 34-35%, and repeated long frames disappeared in both runs. A matching Release Headless control-level allocation probe reduced p50/p95 allocation from 548,648/548,664 B to 37,720/37,736 B per update.

This PR intentionally contains only the library control change; it does not include Demo-page changes or the separate performance-test infrastructure.

### DIFF
--- a/SukiUI/Controls/Gauges/RadialGauge/RadialGauge.cs
+++ b/SukiUI/Controls/Gauges/RadialGauge/RadialGauge.cs
@@ -355,25 +355,6 @@ public class RadialGauge : Panel
         InvalidateArrange();
     }
 
-    private void UpdateSegments()
-    {
-        if (Segments == null || _segmentPaths.Count != Segments.Count)
-        {
-            RebuildSegments();
-            return;
-        }
-
-        for (int i = 0; i < Segments.Count; i++)
-        {
-            var segment = Segments[i];
-            var path = _segmentPaths[i];
-            path.Stroke = new SolidColorBrush(segment.Color);
-            path.StrokeThickness = segment.Thickness;
-        }
-
-        InvalidateVisual();
-    }
-
     private void OnSegmentsChanged(AvaloniaPropertyChangedEventArgs e)
     {
         if (e.OldValue is IList<RadialGaugeSegment> oldCollection)
@@ -417,7 +398,6 @@ public class RadialGauge : Panel
     private void OnSegmentPropertyChanged(object? sender, PropertyChangedEventArgs e)
     {
         InvalidateSegmentGeometry();
-        UpdateSegments();
         InvalidateArrange();
     }
 
@@ -590,14 +570,14 @@ public class RadialGauge : Panel
             _segmentsGeometryDirty = true;
         }
 
-        if (!_segmentsGeometryDirty)
-            return;
-
-        if (Segments == null || _segmentPaths.Count != Segments.Count)
+        if (_segmentPaths.Count != (Segments?.Count ?? 0))
         {
             RebuildSegments();
             return;
         }
+
+        if (Segments == null || !_segmentsGeometryDirty)
+            return;
 
         var center = rect.Center;
         double radius = rect.Width / 2.0;
--- a/SukiUI/Controls/Gauges/RadialGauge/RadialGauge.cs
+++ b/SukiUI/Controls/Gauges/RadialGauge/RadialGauge.cs
@@ -122,6 +122,12 @@ public class RadialGauge : Panel
     private List<Path> _segmentPaths = new();
     private Grid _segmentsGrid = null!;
     private readonly RotateTransform _needleTransform = new(0, 0, 0);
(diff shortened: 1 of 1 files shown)


sn4k3:
Have done some tweaks

1. Trail geometry – build once, mutate in place                                                         
2. GetTrailBrush() – was allocating a LinearGradientBrush + GradientStops + 3 GradientStop every frame. 
  Bigger than the geometry. Now cached, keyed on the matched segment color; rebuilds only when Value      
  crosses a segment boundary.                                                                             
3. Value.ToString("F0") – only reformat when the rendered digits change.
4. introduced SegmentVisual – a private bundle holding each segment's Path, PathGeometry,
  PathFigure, ArcSegment and SolidColorBrush, built once in RebuildSegments, mutated in place by          
  UpdateSegmentsGeometry.
                                                                                                          
```
  ┌────────────────────────┬──────────┬─────────┐                                                         
  │                        │  before  │  after  │                                                         
  ├────────────────────────┼──────────┼─────────┤                                                         
  │ alloc per value update │ 15,567 B │ 1,169 B │                                                         
  └────────────────────────┴──────────┴─────────┘             
```                                            
                                                                                                          
13× reduction



Can you check in your test?

KaranocaVe:
Thanks — I reran this after the latest changes and pushed `348b44e` to preserve the original value-text rendering contract. The first cache implementation used `Math.Round(..., AwayFromZero)`, which changed the visible `Value.ToString("F0")` output at midpoints and for signed zero (for example, `-2.5`, `-0.5`, `0.5`, and `2.5`). The replacement uses `TryFormat("F0")` into stack storage and compares the exact formatted text, so it allocates a string only when the rendered text changes without changing the previous formatting semantics.

I used a real macOS Desktop/Metal session at 2.0 scaling: 24 gauges, 24 segments and 24 ticks per gauge, 90 warm-up frames, then 300 measured value-update frames. Allocation samples cover the complete UI-thread interval between rendered frames, including Avalonia scheduling/layout work, so they are intentionally broader than the local per-method allocation probe.

| Run | `f00ee2f` p95 allocation / frame | Current p95 allocation / frame | `f00ee2f` p95 frame interval | Current p95 frame interval | Missed 60 Hz frames |
| --- | ---: | ---: | ---: | ---: | ---: |
| 1 | 748,296 B (31,179 B/update) | 269,616 B (11,234 B/update) | 16.679 ms | 16.719 ms | 0 -> 0 |
| 2 | 748,056 B (31,169 B/update) | 270,256 B (11,261 B/update) | 16.705 ms | 16.695 ms | 0 -> 0 |

That is a 63.9-64.0% reduction in the complete-frame p95 managed allocation for this dense scenario; the p95 frame interval remains within the display cadence in both versions.

I also verified value-text fidelity for midpoint and signed-zero cases, value-only `SegmentVisual`/geometry reuse, and in-place range, color, thickness, opacity, and active-trail-color updates. `dotnet build Suki.sln -c Release --no-restore -clp:ErrorsOnly` and the PR Native build both pass.

sn4k3:
Is the `Span<char> displayText = stackalloc char[512];` really necessary at that size? Most usages are 0 to 100, but even passing something like download bytes a [64] is ok, if someone is using `double.MaxValue` I think theres a problem already 😅

But looks good to me, is this ready to merge?

KaranocaVe:
Good call — I reduced the stack buffer to 64 chars in `61e04dd`. That covers `F0` output through `1e63` exactly; larger values (including `double.MaxValue`) take the existing `ToString` fallback, so formatting remains correct without reserving the larger buffer on the common path.

The Release solution build and the Native PR build both pass. From my side, this is ready to merge.

PR #645, opened 2026-08-23
perf: avoid duplicate SquishyBehavior animations

KaranocaVe:
## Summary

- keep one routed pointer-event pipeline instead of registering both routed and direct handlers for the same events
- make enable and disable idempotent, and remove the routed handlers when disabled
- keep one reset timer in behavior state, replacing or stopping it on press, release, and disable
- clear the behavior transform and state on disable
- initialize state from attached-property values so XAML property order does not cause null dereferences
- preserve the existing transform calculations and 700 ms spring reset

## Problem

A normal pointer release reached SquishyBehavior through both AddHandler and direct PointerReleased subscription. That launched two independent 16 ms dispatcher timers for the same 700 ms spring animation. Disable removed only the direct subscriptions, leaving routed handlers and animation state installed.

## Performance

Measured with an Avalonia Headless Release probe on .NET 10. The script performs one 55 x 35 pixel drag and release, waits 850 ms, and reports the median of three runs. The baseline is upstream main at df3f7bf; the optimized result is a8e89c9.

| Enable/disable cycles before drag | Transform writes before | Transform writes after | Managed allocation before | Managed allocation after |
|---:|---:|---:|---:|---:|
| 0 | 534 | 270 | 252,096 B | 211,096 B |
| 5 | 534 | 264 | 252,096 B | 207,416 B |
| 20 | 534 | 264 | 252,096 B | 207,584 B |

For the normal no-toggle path, this is a 49.4% reduction in transform property writes and a 16.3% reduction in managed allocation. Repeated enable/disable cycles no longer accumulate work. CPU samples were intentionally excluded because short Headless process samples were noisy.

## Validation

- scripted Headless drag/release benchmark
- Headless regression verifying configuration-before-enable and that disable stops an in-flight reset animation
- dotnet build Suki.sln -c Release

### DIFF
--- a/SukiUI/Animations/SquishyBehavior.cs
+++ b/SukiUI/Animations/SquishyBehavior.cs
@@ -1,4 +1,5 @@
-﻿using Avalonia;
+﻿using System.Diagnostics;
+using Avalonia;
 using Avalonia.Controls;
 using Avalonia.Input;
 using Avalonia.Interactivity;
@@ -17,6 +18,7 @@ namespace SukiUI.Animations
             public ScaleTransform Scale = new(1, 1);
             public SkewTransform Skew = new(0, 0);
             public TranslateTransform Translate = new(0, 0);
+            public DispatcherTimer? ResetTimer;
 
             public double Intensity = 1;
             public double SquishDepth = 1;
@@ -26,7 +28,7 @@ namespace SukiUI.Animations
             public bool ScaleByXY = true;
         }
         
-        private static readonly AttachedProperty<State> StateProperty = AvaloniaProperty.RegisterAttached<Control, State>(
+        private static readonly AttachedProperty<State?> StateProperty = AvaloniaProperty.RegisterAttached<Control, State?>(
             "State",
             typeof(Control)
         );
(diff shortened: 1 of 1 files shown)


sn4k3:
I made this edit:

SquishyBehavior now stores and restores the control’s previous RenderTransform when enabling/disabling, instead of always nulling it out. It also centralizes attached-property-to-state updates, aligns ScaleByXYAxis default behavior, and stops/reset animations when the control detaches to avoid timers mutating visuals after removal.

(Please confirm the correctness of it)

KaranocaVe:
I rechecked the follow-up at `e7641be` with a fresh Release Headless rerun against `df3f7bf`.

Method: a real routed 55 x 35 px drag/release, the 700 ms spring reset, then an 850 ms controlled dispatcher pump. I ran five interleaved process samples on the same machine with .NET 8 and Avalonia 12.1.1; values below are medians.

| Metric | Baseline | Current head | Change |
| --- | ---: | ---: | ---: |
| Transform property writes | 522 | 264 | -49.4% |
| Managed allocation | 246,608 B | 209,928 B | -14.9% |

The lifecycle follow-up is correct: a pre-existing `RenderTransform` is restored after disable (including a repeated disable), and the reset timer is stopped on disable/detach. The `ScaleByXYAxis` default is now consistently `true`, matching the old enabled-state behavior rather than changing the implicit animation path.

`dotnet build Suki.sln -c Release -clp:ErrorsOnly` also passes at the PR head.

KaranocaVe:
ready to merge

PR #646, opened 2026-08-23
perf: release behavior overlays when controls detach

KaranocaVe:
## Summary

- release GlowBehavior and ErrorBehavior popup resources synchronously when their control leaves the visual tree
- stop the per-control dispatcher timer, detach ScrollViewer subscriptions, clear popup content and logical parents, and remove static dictionary entries
- preserve active behavior semantics by recreating the overlay after reattachment
- cancel superseded fades so a stale asynchronous cleanup cannot remove a newly reactivated overlay
- store ErrorBehavior's original opacity with the popup lifecycle instead of in a second static control dictionary

## Problem

Both behaviors used controls as strong keys in static dictionaries. Detaching a control only closed its popup and stopped its timer, leaving the dictionary entry, popup content, logical-parent chain, and subscriptions alive. Repeated navigation therefore retained every detached control subtree.

## Performance

Measured with an Avalonia Headless Release probe on .NET 10. Each scenario repeatedly created and detached batches of 20 controls, disconnected the closed window content, forced full GC, and then inspected the private behavior dictionary plus WeakReferences. The baseline is upstream main at df3f7bf; the optimized result is 3ae759b.

| Behavior | Detached controls | Dictionary entries before | Dictionary entries after | Live WeakReferences before | Live WeakReferences after | Managed heap delta before | Managed heap delta after | Reduction |
|---|---:|---:|---:|---:|---:|---:|---:|---:|
| Glow | 20 | 20 | 0 | 20/20 | 0/20 | 1,591,320 B | 172,712 B | 89.1% |
| Glow | 100 | 100 | 0 | 100/100 | 0/100 | 9,276,400 B | 340,560 B | 96.3% |
| Error | 20 | 20 | 0 | 20/20 | 0/20 | 4,234,640 B | 1,484,312 B | 65.0% |
| Error | 100 | 100 | 0 | 100/100 | 0/100 | 15,807,408 B | 1,627,752 B | 89.7% |

The heap delta is a forced-GC process measurement and includes Headless framework noise; the dictionary and WeakReference results directly verify that detached controls are no longer retained.

## Validation

- Headless detach/GC benchmark for 20 and 100 Glow and Error controls
- Headless lifecycle regression covering detach, active reattach, fast deactivate/reactivate, and final cleanup
- dotnet build Suki.sln -c Release

### DIFF
--- a/SukiUI/Animations/ErrorBehavior.cs
+++ b/SukiUI/Animations/ErrorBehavior.cs
@@ -18,6 +18,13 @@ public static class ErrorBehavior
     private const double ErrorOpacityTarget = 0.1;
 
     private static readonly Dictionary<Control, PopupData> _popups = new();
+
+    /// <summary>
+    /// Popups grouped by the <see cref="ScrollViewer"/> they scroll with, so a scroll event only
+    /// touches the popups that need repositioning and only subscribes once per host.
+    /// </summary>
+    private static readonly Dictionary<ScrollViewer, List<PopupData>> _scrollHosts = new();
+
     public static readonly AttachedProperty<bool> IsActiveProperty =
         AvaloniaProperty.RegisterAttached<Control, bool>(
             "IsActive",
@@ -303,10 +310,7 @@ public static class ErrorBehavior
             timer.Start();
         };
 
-        if (scrollViewer != null)
-        {
-            scrollViewer.ScrollChanged += OnScrollChanged;
-        }
+        RegisterScrollHost(popupData);
 
         _popups[control] = popupData;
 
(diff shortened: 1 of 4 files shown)


sn4k3:
## Changes

Replace per-popup ScrollChanged subscriptions with a shared scroll-host registry (_scrollHosts) so each ScrollViewer is subscribed only once, preventing duplicate events and handler accumulation.

Also fixes related bugs:
- Capture live opacity before cancelling in-flight animations to prevent no-op fade-outs
- Use SetCurrentValue instead of direct Opacity assignment to avoid stomping animated values
- Replace anonymous DetachedFromVisualTree closures with static handlers in VisibilityBehavior to prevent control leaks
- Guard CancellationTokenSource access with Interlocked.Exchange to avoid ObjectDisposedException
- Fully release LoadingBehavior overlays on detach and rebuild on reattach
- Extract ReleasePopup/SubscribeToLifecycle helpers for consistency

Please check

kikipoulet:
What a time to be alive

KaranocaVe:
Rechecked the latest follow-up locally at `9f9b0d2`.

- `dotnet build Suki.sln -c Release` passes.
- In a Headless scenario with 64 active controls of each behavior inside one `ScrollViewer`, Loading, Glow, and Error each kept 64 popup records under exactly one scroll-host entry. After deactivation, both their popup and scroll-host registries were empty.
- Rapid deactivate/reactivate followed by a final deactivation drained Glow and Error correctly; Error also restored the original `0.8` opacity.
- VisibilityBehavior removed its dimension-cache entry on detach. LoadingBehavior released its popup on detach and rebuilt it on reattach while correctly blocking input.

The follow-up looks sound to me.

One separate pre-existing case remains: a fast `LoadingBehavior.IsBusy: true → false → true` transition can still end with `IsBusy=true`, no popup, and input enabled. I reproduced that on both `df3f7bf` and this head, so it is not introduced by this follow-up and should be handled independently.

sn4k3:
Is this ready to merge?

KaranocaVe:
> Is this ready to merge?

yeah

PR #650, opened 2026-09-13
build(nuget): set PackageLicenseExpression to MIT

BrycensRanch:
NuGet packages should declare their license via an SPDX expression so that consumers can automatically analyze license compliance. The previous package had no license metadata, which caused warnings in tools like NuGetPackageExplorer.

Fixes errors like:
SnapX/Directory.Packages.props(66,5): warning Proj0500: The SukiUI (7.0.1) package is shipped without an explicitly defined license (https://dotnet-project-file-analyzers.github.io/rules/Proj0500.html)


sn4k3:
Thanks, I've consolidate and modernize the csproj, by using .props.
Now it propagates to all libraries and not just the main.

PR #657, opened 2026-09-18
Extract optional ColorPicker and ConditionalXAML packages

### LINKED ISSUE #656: [Feat] Move Avalonia ColorPicker support to an optional package

- [x] I have searched for duplicate issues.

Move the `Avalonia.Controls.ColorPicker` dependency and SukiUI ColorPicker theme resources out of the core `SukiUI` package into a separately published `SukiUI.ColorPicker` package.

The ColorPicker controls are optional, but the dependency and custom theme are currently included for every SukiUI consumer. Keeping them in an optional package will reduce the core dependency surface and avoid imposing ColorPicker trim/AOT analysis on applications that do not use these controls, while preserving the current styling for applications that opt in.

The demo application and release workflows should be updated to reference and publish the optional package. Installation documentation should explain how to install and include its styles, following the existing `SukiUI.DataGrid` package pattern.

### Package Version

main

CiaranDall:
## Summary

- remove `Avalonia.Controls.ColorPicker` from the core `SukiUI` package
- add a packable `SukiUI.ColorPicker` project containing the ColorPicker dependency and SukiUI ColorPicker theme
- remove `System.Linq.Dynamic.Core` from core and add a packable `SukiUI.ConditionalXAML` project containing the public `If` and `InlineSharp` helpers
- update the demo to opt in explicitly, add release workflow packaging, and document installation

## Motivation

Both ColorPicker and ConditionalXAML are optional functionality. Keeping them out of the core package reduces the dependency surface for applications that do not use them. The ConditionalXAML helpers retain their original public namespace through the optional package.

The extraction follows the existing `SukiUI.DataGrid` package pattern. Native AOT demo handling is tracked separately.

## Validation

- `dotnet restore Suki.sln`
- `dotnet build Suki.sln -c Release --no-restore` (0 errors; existing warnings remain)
- packed `SukiUI` and `SukiUI.ColorPicker` and verified optional dependencies
- verified the demo references `SukiUI.ConditionalXAML` for the helper namespace

Closes #656


sn4k3:
Thank you :)
Title and description talks about `SukiUI.ConditionalXAML` but not found on commit?

CiaranDall:
Fixed - `SukiUI.ConditionalXAML` is now included in the PR branch.

sn4k3:
Thanks, while reviewing... Is it possible to rework ConditionalXAML to get rid of dynamic linq dependency and at same time make it work and compatible with AoT?

CiaranDall:
It is possible, but not as part of this extraction. ConditionalXAML currently supports arbitrary string expressions, including property access, arithmetic, ternaries, and static types; preserving that API without Dynamic LINQ would still require runtime parsing/reflection. An AOT-safe redesign would need a restricted compile-time DSL/source generator or regular compiled bindings/ViewModel properties, so I would handle it in a separate design and PR.

PR #660, opened 2026-09-19
MacOS Window Chrome

LuxuryPanda:
## Summary

Added a macOS title bar chrome to `SukiWindow`, traffic light window controls on the left edge, with the platform's own full screen behaviour.

### Showcase

image

Hovered:
image

The icons could be improved further to exactly match the native ones, but IMO are fine for this first implementation.

## What this adds

A `WindowChromeMode` property (`Auto`, `Windows`, `MacOS`) selects the control style, defaulting to the platform the application runs on. In macOS mode the window controls dock to the leading edge as round traffic lights, their glyphs appear only while the pointer is over the group, custom title bar controls move to the trailing edge where the platform puts application controls, and double-tapping the title bar zooms the window. `MacWindowNativeActions` routes zoom and full screen through `NSWindow` so the window joins a real macOS full screen space rather than merely filling the display.

The green traffic light is the full screen control, matching the platform, where the zoom box appears only with Option held. Its glyph is a pair of solid triangles on the NW/SE diagonal, pointing outward to enter and inward to leave.

## Details worth a reviewer's attention

**Full screen state reporting.** macOS reports leaving full screen as `FullScreen -> Minimized -> Normal`. The `Minimized` early return in `OnWindowStateChanged` would skip the title bar restore, and because `MaxWidthScreenRatio` defaults to `NaN` nothing clamps the result either, so the window returned sized to the full screen with its title bar hidden. The pre-full-screen bounds are captured on entry and both restores run once the window settles, whichever sequence macOS reports.

**Animation cancellation.** `AnimateAsync` uses `FillMode.Forward` and discarded its `CancellationTokenSource`, so toggling full screen quickly let a stale hide animation land after the show animation and pin `ScaleY` to zero while `IsTitleBarVisible` was still `true`, collapsing the title bar for good. `AnimateAsync` now accepts a `CancellationToken` and `SukiWindow` cancels the in-flight animation before starting the next one, which replaces the previous version-counter guard.

## Testing

Verified by hand on macOS: repeated full screen cycles, fast and slow, returning the window to its original size and position with the title bar intact, and the traffic light glyph flipping between the outward and inward arrows.

## AI Disclosure

This has been implemented with the help of AI, I've tried to scope down the changes to the bare minimum, and polished this quite heavily to reach the closest look to native chrome.


sn4k3:
Looks good to me, I always thought that we need this kind of "native" bar to mimic macOS style.
I don't have a mac to fully test this out, is the behaviour the same if I test on windows?

LuxuryPanda:
> Looks good to me, I always thought that we need this kind of "native" bar to mimic macOS style. I don't have a mac to fully test this out, is the behaviour the same if I test on windows?

Yes, the behaviour is pretty much the same, aside from the fullscreen green button which will just maximise the screen on Windows

sn4k3:
> Yes, the behaviour is pretty much the same, aside from the fullscreen green button which will just maximise the screen on Windows

Can a demo be added in the dialog section to open a window with macOS chrome? This feature is interesting to have in the catalog.

LuxuryPanda:
> Can a demo be added in the dialog section to open a window with macOS chrome? This feature is interesting to have in the catalog.

Sure thing, adding it 🚧

LuxuryPanda:
Hi @sn4k3, it's done, please check it out.

Demo buttons:
image

Dialog with MacOS chrome:
image

Dialog with Windows chrome:
image

sn4k3:
Added some fixes, please test

LuxuryPanda:
> Added some fixes, please test

Looks good, haven't seen any issue

